Issue #5, opened 2023-10-24
Create Contact Us page


mradamcox:
See designs in Figma. This will include a form that will be integrated with Netlify's form integration service.

PR #11, opened 2023-10-25
feat: initial content creation for contact us page

mradamcox:
Initial creation of "contact us" page for #5. This includes the main elements, but still lacks

- A unified container spacing - this should be taken from other pages that are still in progress.
- The footer - this should be inherited from a shared component, but currently the footer is only hard-coded into the homepage
- The "Clear" button for the form


mradamcox:
Merging now, the footer has been added, still more spacing/styling to do.

Issue #76, opened 2024-01-30
Initial draft of results list


mradamcox:
Create list of result items displayed through the card from #75. This list should only include "parent" records, i.e. those without a `dct_isVersionOf_sm` record. For now, this can be place in a left-hand panel/sidebar in the search page (we can configure later).

pengyin-shan:
Close this issue since its PR has been merged to `discovery_app` branch

Issue #75, opened 2024-01-30
Initial draft of discovery app parent resource card


mradamcox:
We will have a results list in the discovery app, and this ticket is to develop an "result item card" component that will be used for each returned result. This card will only be used for "parent" records, i.e. those that don't have a `dct_isVersionOf_sm` value. Child records (those that do have an `dct_isVersionOf_sm`) for each parent will be referenced within this card.

To start, this card should have the following elements from a given item:

### {`dct_title_s`}
_{`dct_creator_sm`}_
{`dct_description_sm`} (truncated to 25-30 words)
Years: {for each **child record**, show child record `dct_issued_s`}

**Important** the objective of #74 is to translate the Solr document such that instead of using the harder to predict `dct_title_s` attributes, we'll use `title`, `creator`, `description`, etc. This means that exact reference will be something like `item.title.value`, instead of what is presented above. If necessary, you should be able to use `item.meta.multiple` to predict values that are arrays.

pengyin-shan:
Close this issue since its PR has been merged to `discovery_app` branch. After the designer gives the design for the parent object, a new issue will be created.

PR #81, opened 2024-02-05
#74, #75, #76: Change raw Solr object to local Solr object and create Parent Object for side bar

### LINKED ISSUE #74: Create helpers.js to load metadata schema and translate Solr records
Use `discovery_app` branch.

We should add a function (or two) to pages/search/_metadata/helpers.js to do the following:

1. Get full metadata schema

Load both metadata schema files in that directory, combine and return them in a single object.

2. Translate Solr records for easier access in templates

Take a json object input from Solr, and use the lookup produced above to output and object with a format that uses the more readable id as key and includes a "meta" property with the field-specific attributes. For example:

input solr document json (partial):

```
{ 
  dct_title_s: "CDC Social Vulnerability Index (SVI) 2000"
}
```

output object:
```
{
  title: {
    meta: {
      "label": "Title",
      "uri": "dct_title_s",
      "data_type": "string",
      "multiple": false,
      "obligation": "required",
      "controlled": false,
      "controlled_options": [],
      "schema": "Aardvark",
      "display_group": "Descriptive",
      "slug": "title",
      "widget": "text-simple.html"
    },
    value: "CDC Social Vulnerability Index (SVI) 2000"    
  }  
}
```

It's likely that we won't need all of these meta properties, but for now we should begin by making everything available.

pengyin-shan:
This PR is intended to resolve #74, #75 and #76. 

This PR defines the `SolrObject` interface (#74), which is the basic object type that all Solr raw objects should be transferred to before future processes in the front end. Then based on different user requirements we will transfer the SolrObject to other types of objects.

This PR also includes `SolrParent` interface (#75), which is the parent record type that is used in the sidebar of the Discovery page. Currently, they are displayed as accordions as a placeholder to show Parent Object’s data (#76), but we can update the visual components to what designers want later.

### How to Test

1. Run `npm i` to install the MUI icon library for Accordion
2. Run `npm run dev` and navigate to http://localhost:3000/search
3. You should see the parent list accordions. Expand any accordion and you should see creator information, description, and years. 
Screenshot 2024-02-05 at 9 05 18 AM

Note that:
- Description of the `SolrParent` is the first sentence of the description of `SolrObject` 
- Years are generated based on child records using the `is_version_of` attribute, with *duplication removed*. If we want to see duplicate years, I can change the Set back to Array for the `years` attribute of the `SolrParent` interface.
- **When defining the parent object, the basic idea is that as long as the object has no `is_version_of`, this object is a parent object even though there could be no child**. If we require a parent object **must** has years (i.e. children), then I'll update the code accordingly


pengyin-shan:
> @pengyin-shan this really great, a perfect place for us to start. Two things: 1) would it be possible to replace `rawSolrMetdataSwitch` with a function that dynamically loads the two JSON schema files, instead of hard-coding these variable names? I don't think it would impact performance (as NextJS is pre-rendering these pages) but I may have that wrong. 2) Because anything under `pages` becomes a publicly addressable route, could you move or rename `parentList.tsx`? (I don't think we'll need `/search/parentList`). Here's [more on NextJS routing)[https://nextjs.org/docs/app/building-your-application/routing/colocation), and if you have other ideas for restructuring, I'd say go for it. Thank you, this is looking really good!

@mradamcox, here are the updates I made:

1. I modified the code to use aardvark_schema.json and sdohplace_schema.json for defining the meta fields of SolrObject. The attributes id, title, metadata_version, modified, access_rights, and resource_class remain unchanged as they are the required attributes for the interface, the same as SolrParent. The content of the two JSON files is loaded into solrObjects.js, where the method to initialize raw Solr objects is retained.

2. Sorry that I overlooked the fact that Next.js follows convention-based routing...I moved the parentList component to the components folder instead.

mradamcox (approved):
Splendid, this looks good. We'll definitely have hard-coded identifiers for the display objects like this, but loading the entire schema files at once like this will be useful down the road I think.

PR #84, opened 2024-02-13
Feature/#83: build the basic search feature for search page

pengyin-shan:
This PR includes a basic search box and backend that we can do further extension later.

### How to Test

1. Run the app and navigate to http://localhost:3000/search.
2. You should see an unstyled basic search box at the bottom of the page. We can restyle it later.
Type anything in the search box:
- If there's any prefix match on the title, you should see a popup appear (Try "new" and you should see "new one" popup):
Screenshot 2024-02-13 at 11 04 22 AM

- If you are typing stuff that won't have a *prefix match*, you won't see anything pop up

4. Click "new one" in the popup. You should see the exact object returned (we can choose to expand the object later; for now, I'm just outputting basic attributes of it).

Screenshot 2024-02-13 at 11 34 04 AM

6. Don't type anything in the search box, just click it. You should see all results available because this means the user is using the prefix `*`. We can disable this function later if we don't need it.
Screenshot 2024-02-13 at 11 34 31 AM

 7. Type 'US' in the search box (representing random user input) and click 'Search', you see should multiple results returned because the general query `https://solr.sdohplace.org/solr/blacklight-core/select?q=US` is called:
Screenshot 2024-02-13 at 11 33 12 AM

 8. Type 'US' in the search box again and select one result from the pop up, you should see only one result returned based on your selection:

Screenshot 2024-02-13 at 11 33 33 AM

 ### Future Actions
 
1. I'd like to know what the query syntax is for autocomplete in GeoBlacklight (sorry, I'm not very familiar with it). Currently, I just use prefix search for the title field. After I learn the detailed syntax, I'll replace it.
2. Currently, I only use user input to search, but I have defined other GeoBlacklight fields in the interface. After we finalize other criteria for searching, they will be used.
3. I have listed a variety of query methods, but currently, only general query and prefix query are used in this PR. Once we discuss the exact query features we need, we can call these methods.

### DIFF
--- a/meta/helper/solrObjects.ts
+++ b/meta/helper/solrObjects.ts
@@ -1,5 +1,5 @@
 import { SolrObject } from "../interface/SolrObject";
-import { SolrParent } from "../interface/SolrParent";
+import { SolrParent } from '../interface/SolrParent';
 import aardvark_json from "../../src/pages/search/_metadata/aardvark_schema.json";
 import sdoh_json from "../../src/pages/search/_metadata/sdohplace_schema.json";
 import { AardvarkSdohSchemaMatch, findFirstSentence } from "./util";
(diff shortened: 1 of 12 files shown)


mradamcox (changes requested):
@pengyin-shan This is looking really good, but let's try changing to Solr's `/suggest` endpoint for autocomplete options. So, can you switch things so that whenever input is changed in the box, its value is passed to `(solr url)/suggest?q=<input>`? You'll find that endpoint isn't case-sentive, so just `(solr url)/suggest?q=a` will return a lot of suggestions. (Note: I recognize that the endpoint returns facet values that aren't currently supported by the search, but that's ok for now.)

pengyin-shan:
> @pengyin-shan This is looking really good, but let's try changing to Solr's `/suggest` endpoint for autocomplete options. So, can you switch things so that whenever input is changed in the box, its value is passed to `(solr url)/suggest?q=<input>`? You'll find that endpoint isn't case-sentive, so just `(solr url)/suggest?q=a` will return a lot of suggestions. (Note: I recognize that the endpoint returns facet values that aren't currently supported by the search, but that's ok for now.)

Hello @mradamcox, based on our discussion last Friday, I have revised most parts of the code to reflect what we will need and prepare for adding filters. Please test following these steps:

1. Run the app and navigate to http://localhost:3000/search.
2. You should see the search box now move to the top of the page and combine it with the parent list (please ignore the styling).
3. Start typing `n`, and you should see suggestions pop up, similar to what you will see in our Geoblacklight space.
4. Select `Nampa, Idaho`, and you should see the corrected parent item displayed as an accordion (we can change this component later as we like).
5. Expand the parent object returned from step 4. As proof of the search box's result, if you search for 'Nampa, Idaho' on the page, you should see this item appear as part of the metadata:
Screenshot 2024-02-19 at 1 33 05 PM

6. Clear the input box and try `unemployment ci high` by tying `u` this time. You shouldn't see anything in the parent list.
7. This is because the returned item is a child element. As we discussed last Friday, for now we don't show them, but if you check the console, you can still see the search successfully returned the result:
Screenshot 2024-02-19 at 1 46 15 PM

8. Search for something that won't have a suggestion pop-up and submit the search form. Now you should see nothing returned. Feel free to also try other searches.

I hope these changes fit the purpose we discussed on Friday. Please let me know if you have any questions. Thank you!

mradamcox:
Splendid, yes this looks like just what we discussed. I will merge now. Thank you!

Issue #85, opened 2024-02-15
Update Logo and activate link to Community Toolkit


mradamcox:
We now have a logo to replace the existing draft logo. Additionally, the toolkit is published now at https://toolkit.sdohplace.org. 

- [x] update logo and favicon
- [x] update link to Community Toolkit at the top and bottom of the home page.
- [x] anything else? Maybe there is another place on the site that should be linked to the toolkit, just give a quick scan around.

Logo/favicon assets will be supplied via google drive.

PR #89, opened 2024-02-22
[WIP] #85: Search Filters

pengyin-shan:
Draft PR, Still working in process.

Done:
1. Basic filter in and out feature
2. [Search Enhancement] allow user to search items without following the dropdown hit and still see parent objects.
3. [Search Enhancement] multi-term seraches

Next steps:
1. add number changing function to filters when users performing searching
2. add other attributes to filters 
3. add 'more' to spatial resolution
4. probably some code refine


pengyin-shan:
UPDATE: 
1. Allow filter's number changes with results changing. Revise some interfaces and methods.
2. after checking the initial efficiency of the purely BROWSER side of the filter, the performance will be too slow if the user enables too many filters. Thus I'll switch to the query methods to improve the performance and reduce code complexity.

mradamcox:
>     2. after checking the initial efficiency of the purely BROWSER side of the filter, the performance will be too slow if the user enables too many filters. Thus I'll switch to the query methods to improve the performance and reduce code complexity.

Great, this is definitely the right move here. Thank you for recognizing that!

pengyin-shan:
Update the filtering method to query. Next optimization steps:
1. Instead of rendering all attributes of the filter once, separate filter rendering to each attribute. When the user clicks filter, only update the corresponding attribute instead of the whole filter list
2. Instead of generating a parent object from the child object, filter out the parent object for the returned result.

pengyin-shan:
This PR is ready for review. @mradamcox please see my top comment for updates

mradamcox (approved):
Outstanding, this is coming along really well, and the page performance within the deploy preview is snappy. I appreciate the modular approach and refinements you are making as you develop the internal search API. One thing we'll definitely need is a "clear filters" button like you have described in #102, but no huge rush on that. I will merge this PR now.

Issue #102, opened 2024-02-27
Add a 'Clear out search result' button.


pengyin-shan:

pengyin-shan:
close this issue as the code has been merged

PR #108, opened 2024-02-29
#102: Add a button to start over user search

pengyin-shan:
This PR is for issue #102, intending to perform the same functionality as the "Start Over" function currently available in the GeoBlacklight platform:
Screenshot 2024-02-29 at 11 27 31 AM

I also added the feature that when the user clicks the 'X' on the input button, we should treat it the same as if the user wants to start over (i.e., clear out all filters and reset both results and the input box).

## How to Test

1. Run the app and input "av" in the input box, then click "Search":
Screenshot 2024-02-29 at 11 30 37 AM

2. You should see all results based on the suggestions of "av" returned. Select any filter to apply it.

3. You should see a "Start Over" button appear. Click it.

4. Now all filters and results should be reset to their original status, and the input box should be cleared.

5. Run the app and input "av" in the input box, then click "Search" again. You can perform some filtering actions.

6. This time, click the 'x' button in the pop-up in the input box:
Screenshot 2024-02-29 at 11 34 45 AM

7. The same reset effect should apply to the page. All filters, the input box, and results should be reset.

### DIFF
--- a/src/components/search/searchArea.tsx
+++ b/src/components/search/searchArea.tsx
@@ -20,10 +20,7 @@ import { SearchObject } from "./interface/SearchObject";
 import SolrQueryBuilder from "./helper/SolrQueryBuilder";
 import SuggestedResult from "./helper/SuggestedResultBuilder";
 import ParentList from "./parentList";
-import {
-	filterParentList,
-	generateSolrParentList,
-} from "meta/helper/solrObjects";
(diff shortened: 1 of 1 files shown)


pengyin-shan:
FYI, you may need to checkout this branch locally and use production endpoint to test

mradamcox (approved):
Looks good!

PR #120, opened 2024-03-04
77 integrate search map

mradamcox:
Drafting a PR that will combine the map interface with the in-progress search page. The objective is to have a generally functional first draft that can illustrate the capabilities we will be able to implement once we have a solid idea of the desired design.


mradamcox:
I've combined the elements into a single page, and added a Mui "Drawer" to show details for individual records once they are selected: Here's how things are looking
image
image

mradamcox:
I'm going to merge this now, so @pengyin-shan you can begin working on some further refinements that I'll describe in a few tickets.

Issue #121, opened 2024-03-05
Add close button to record Drawer


mradamcox:
The drawer that pops up with more info for each record on the search page needs to have a button to close it (clicking outside of the drawer area works but isn't intuitive enough.

In the top right corner of the drawer a button should be added that says "back to search results" which will close the drawer when clicked.

pengyin-shan:
Close this issue as the code has been merged

Issue #115, opened 2024-03-04
refactor: Aardvark & SDOH schema to single object


mradamcox:
In a few places, both the Aardvark schema and our extra SDOH schema are loaded as individual JSON objects, and then passed into functions like`findSolrAttributes` for interrogation. It would be better to have a single helper method that loaded and combined any `.json` files inside of `src/pages/search/_metadata` and returned a combined object. Then functions like `findSolrAttributes` would only take a single `schema: {}` argument, instead of the current two `aardvark_json: {}, sdoh_json: {}` arguments.

mradamcox:
Yes, and the top-level ids will be unique across all files. This isn't urgent at all, more of a small detail, just wanted to write it down while I was thinking about it.

pengyin-shan:
Ok, I'll work on this

pengyin-shan:
Close this issue as the code has been merged

PR #127, opened 2024-03-07
Issue 115 schema to single object

pengyin-shan:
This PR is for #115 and #121 . It will combine all JSONs in the `meta` folder into one JSON to use and add a 'Back to Result' button in the extended drawer.

### How to Test

1. Run the app and navigate to the search page
2. Perform searches and run filters, nothing should change in terms of functions
3. Open http://localhost:3000/api/metadata and you should see the combined JSON file being used
4. Open a drawer by clicking the 'More Info' button from one of the result
5. You should see a 'Back to Results' button in the drawer
6. Click the Back to Results' button. You should see the drawer closed


mradamcox (changes requested):
@pengyin, this is great, just a couple of things. 1) please change to /api/schema instead of /api/metadata, and 2) could you update your local indentation settings to match the existing settings? I noticed a couple of PRs ago that you have a different setting from the rest of the code base. If you think we should change our indentation permanently, we can do that all at once in a separate commit. It would be good to avoid these large diffs like you can see in this PR.

pengyin-shan:
> @pengyin, this is great, just a couple of things. 1) please change to /api/schema instead of /api/metadata, and 2) could you update your local indentation settings to match the existing settings? I noticed a couple of PRs ago that you have a different setting from the rest of the code base. If you think we should change our indentation permanently, we can do that all at once in a separate commit. It would be good to avoid these large diffs like you can see in this PR.

Sure, I updated the code 

> @pengyin, this is great, just a couple of things. 1) please change to /api/schema instead of /api/metadata, and 2) could you update your local indentation settings to match the existing settings? I noticed a couple of PRs ago that you have a different setting from the rest of the code base. If you think we should change our indentation permanently, we can do that all at once in a separate commit. It would be good to avoid these large diffs like you can see in this PR.

Hi @mradamcox , I added a `.editorconfig` for IDE to refer and will use this to avoid my listing error in the future. The new endpoint has been updated to `/api/schema`. Thank you!

mradamcox:
Great. Ready to merge this now.

Issue #130, opened 2024-03-08
Style updates that needs confirmation/resource first


pengyin-shan:
This PR includes style changes that need the resource to be handled first. It is on hold and will be updated on the go as resources are confirmed:

- [x] Change the SVGs of the background arrows on both landing page and footer using the new svgs designer provivded 
- [x] Add hidden text to Etc. at home page and change it to the visible link when the link is ready (since the design has been determined)
- [x] The designer will talk to the team to confirm changing the link of the toolkit logo to the main site
- [x] Update the footer logo after the designer provides it
- [x] Update the footer to include a third column after the designer provides design
- [x] Replace blog post image after designer provides it
- [x] Add two new logos after the designer provides them

pengyin-shan:
More requirements on March 14 

- [ ] Add footer to News page
- [ ] Make Brought to You two columns instead of one column after two more organization

PR #140, opened 2024-03-15
#130: Update styles after designers created new resources

pengyin-shan:
This PR is for #130, which includes the rest of the styling requests from designers after the designer hands the required files. *Note that the MUI Icon library has been added. we can remove it if designers want to create a new icon for it.*

### Tasks and How to Test
**Test methods are _italic_**
- [x]  Change the SVGs of the background arrows on both the landing page and footer using the new SVGS designer provided: _Go to the home page. The background lines (with salmon pink and purple colors) now have rounded edges instead of sharp edges as on the production site._
- [x]  Add hidden text to Etc. at the home page and change it to the visible link when the link is ready (since the design has been determined): _This is currently invisible because we don't have the link yet. If you check card.tsx, you can find the code I added to display content under "Etc" if the link is provided_
- [x]  Update the footer logo after the designer provides it: _Go to the title section of the home page and the footer section. You should see that both logos have been updated. _
- [x]  Update the footer to include a third column after the designer provides design: _Go to the footer section. You should see that a third column has been added to include a button for the News page, matching the following design.:_
Footer 3rd Column
- [x]  Add footer to News page: _Go to News page, you should see that the page now has a footer. Post pages also got the footer added._
- [x] Add two new logos after the designer provides them: _Go to the "Brought to You" section of the home page. You should see two more logos added. Note that the SIEBEL logo is in PNG format for now. The designer will update it once he gets the SVG file_
- [x] Make Brought to You two columns instead of one column after two more organization: _open the mobile view of the site and navigate to the "Brought to You" section of the home page. You should see logos are aligned in two columns instead of one column as the designer requested_


mradamcox (approved):
@pengyin-shan Great, all of these changes look really good. Thanks for the thorough description in the PR. Adding the MUI Icon library is fine, we may need some of those icons in the map search interface as well anyway.

Issue #143, opened 2024-03-25
[Style Update] Change font-size of page introductions


pengyin-shan:
Based on the call with designer, for all pages that have introduction text including home page, change font-size to xl

mradamcox:
All set.

PR #144, opened 2024-03-25
#143: Change font-size of page introductions

pengyin-shan:
This PR is for #143 . Based on call with designer, reduce the introduction font sizes on every page (except News) to `xl`.


mradamcox (approved):
looks good

Issue #147, opened 2024-03-26
Update map layers based on Spatial Resolution


mradamcox:
Once we have a connection between the map and search results/filters, the first thing we need to do is link the Spatial Resolution Filter. When a Spatial Resolution value is selected/deselected, the corresponding layer should be added/removed from the map.

One other change this ticket should include is to update the default map layers. Currently, all map layers are added, but we only want the **state** and **county** layers to be present on initial load.

mradamcox:
Closing now, as this is generally handled. I expect we'll tweak the behavior in the future.

Issue #141, opened 2024-03-20
Connect the filter to map


pengyin-shan:

mradamcox:
Providing some more info about this now. We need to be manipulating the map configuration: what layers are visible, what features are highlighted (this is done by updating a layer's style), zoom extent, etc, via the updated search results.

The best case implementation has map object itself is present and methods can be called on it directly when the search results update. Some ways to achieve this would be:

- Figure out a way to pass a new variable through to the map component such that the `mapRef` object within the component performs an action without needing to re-render the whole component.
- Bring the construction of the map to the top-level module, search/index.tsx, and then manipulate it there
- Same as above, but utilize stateless functions in a helper module that take, for example, a layer id and `mapRef` as arguments and then manipulate the map object within that function.

I consider these options, or something along those lines, preferable alternatives to re-rendering the entire map component with a new set of configurations whenever results are updated. This is because I expect that re-rendering would disrupt in-progress panning and zooming, cause unecessary tile loads, and generally have unexpected side-effects. I may be wrong about this, but I think we need to try the other approaches first because they will provide us the greatest flexibility down the road.

pengyin-shan:
Closing this for now and we will discuss more detail after we make directional decisions

PR #149, opened 2024-04-09
#141 and #142

pengyin-shan:
This pull request addresses #141 and #147 but is in the initial stage and could change direction based on further design decisions.

### How to Test
1. Open the search page. You should see the Spatial Resolution filter removed from the filter list and now at the top of the left panel.

2. During the initialization stage, only state and county layers have been applied to the map. (The blue dots are my test on spatial coverage locations. They are the counties of the City Health Dashboard. I leave the code in case we need this in the future, but they don't belong to #141 and #147).

3. Play around with the Spatial Resolution filter. You should see the layers change based on your selection, but the state and county layers are always there.

4. Perform other random search and filter actions. The Spatial Resolution layer should not be affected by these operations. (i.e., it is a distinct control of the map)

NOTE:

1. I think this matches our plan for the map behavior, but please let me know if not, and I will modify the function accordingly. I think I should probably end here for now and resume after we make more decisions on map behavior.
2. There is a lot of commented-out code and some code that is currently not used. I still keep them there because there's a probability we will need them in the future, as we are still in the decision-making process for the map.


mradamcox:
I'll look at this locally as soon as I can, but for me in the deploy preview the [search page](https://deploy-preview-149--cheerful-treacle-913a24.netlify.app/search) begins to render and then returns an error page with

```
Application error: a client-side exception has occurred (see the browser console for more information).
```

The console shows

```
Uncaught (in promise) TypeError: this.map.style is undefined
```

(firefox and chrome)

pengyin-shan:
> I'll look at this locally as soon as I can, but for me in the deploy preview the [search page](https://deploy-preview-149--cheerful-treacle-913a24.netlify.app/search) begins to render and then returns an error page with
> 
> ```
> Application error: a client-side exception has occurred (see the browser console for more information).
> ```
> 
> The console shows
> 
> ```
> Uncaught (in promise) TypeError: this.map.style is undefined
> ```
> 
> (firefox and chrome)

Ops probably some building errors are going on...thank you for letting me know and I'll fix it

mradamcox:
FWIW the build and deployment process does run without error on Netlify.

pengyin-shan:
> FWIW the build and deployment process does run without error on Netlify.

yes, and my local version works fine 😓 I think it is because the JSON file that contains data is not in the correct folder, thus the API cannot read data correctly, working on the fix now

mradamcox (approved):
This is good progress, and definitely good enough for now. One thing that will need to change is that when a spatial category is selected, then all of the map layers are removed and re-added, instead of only the addition of that single new layer. Something to fix down the road, I agree that we'll revisit this only after getting more design input.

pengyin-shan:
> This is good progress, and definitely good enough for now. One thing that will need to change is that when a spatial category is selected, then all of the map layers are removed and re-added, instead of only the addition of that single new layer. Something to fix down the road, I agree that we'll revisit this only after getting more design input.

That sounds good, once we see more design, let's discuss this and I will modify the special resolution behavior accordingly

Issue #156, opened 2024-05-01
Remove the 3rd logo on main page


pengyin-shan:
Based on the conversation with the designer, remove The UChicago  Center for Spatial Data Science Logo from "brought to you"

mradamcox:
Completed and merged to production.

PR #157, opened 2024-05-01
#156: Remove UChicago logo from homepage and add .env to .gitignore

pengyin-shan:
This PR is for #156. Based on the designer's request, I removed the UChicago logo from the 'brought to you by' section on the homepage. Additionally, I noticed that we don't have .env in .gitignore, so I added it.

### How to Test

Run the app and navigate to the 'brought to you by' section on the homepage. You should see that the UChicago logo has been removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ yarn-debug.log*
 yarn-error.log*
 
 # local env files
+.env
 .env.local
 .env.development.local
 .env.test.local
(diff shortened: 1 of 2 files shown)


mradamcox (approved):
Looks good. `.env` has only been utilized so far in the search app, which isn't merged to main, that's why it's not gitignored here yet. No problem to add this here though.

Issue #164, opened 2024-06-10
Change the map point color to purple


pengyin-shan:
Based on the designer's request, change the points on the discovery map to #7E1CC4. Also modify the color of the Spacial Resolution Selection to make it match the Search icon

PR #165, opened 2024-06-10
#164 change the point color on map to purple

pengyin-shan:
This PR is for issue #164. Based on the designer's request, temporarily change the points on the discovery map to #7E1CC4. Also, modify the color of the Spatial Resolution Selection to match the Search icon.

I put temporary color codes here. In the future, these will be modified based on the designer's final design

### How to Test

1. Run the app and navigate to the search page.
2. Verify that the map pin color has been changed to purple and the background color of the Spatial Resolution Selection has been changed to MUI blue.


mradamcox (approved):
Good.

PR #168, opened 2024-06-12
Remove the purple dots and allow more weights to be shown in suggestion

pengyin-shan:
As we discussed today, I temporarily uncommented the purple dot code to hide the dots on the map (we can still re-enable them if needed). Also noticed that the weight of the suggestion shouldn't be limited to 1, so this PR will fix that.


mradamcox (approved):
I've added a few commits to this PR of work I was in the middle of already when you made it.

Issue #173, opened 2024-06-28
Create component for full item details section


mradamcox:
Create design according to mockups in Figma. More detailed explanation in Notion.

Issue #172, opened 2024-06-28
Create component for search results item


mradamcox:
Create design according to mockups in Figma. More detailed explanation in Notion.

PR #175, opened 2024-07-01
#172: add result list item component

pengyin-shan:
This PR is for #172. I created the component to fit the container's size (in the code, I made the example container 499px to simulate the design). We don't have the mobile design, but I aligned the elements vertically for small screens.

Note that we will need new icons. Currently, I just pass an existing icon on the home page to the test component.

-----
### How to Test

- Navigate to the search page; an example result component should be on the top.
Screenshot 2024-07-01 at 3 17 18 PM

- Change to mobile view and check the mobile layout of this component.
Screenshot 2024-07-01 at 3 17 44 PM

-----
During the development, I found that it would be better to define the exact objects under 'meta' field for the SolrObject. 

Originally I have 
``` javascript
export interface SolrObject {
  id: string;
  title: string;
  ...
  meta: { [key: string]: string | string[] };
  ...
}
```
but `string | string[]` type can cause some problems when passing these meta fields as attributes of the component, so I will work on #176 

After this ticket, I will work on #173.

### DIFF
--- a/src/components/search/iconText.tsx
+++ b/src/components/search/iconText.tsx
@@ -1,0 +1,36 @@
+import Image from "next/image";
+
+interface Props {
+  svgIcon: any;
+  label: string;
+  labelClass: string;
+  labelColor: string;
+  roundBackground: boolean;
+}
+const IconText: React.FC<Props> = ({
+  svgIcon,
+  label,
+  labelClass,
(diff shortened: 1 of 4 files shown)


mradamcox (approved):
Sounds good to me. This is a good start for the item details.

Issue #176, opened 2024-07-01
 Add detail attributes to the meta field of SolrObject


pengyin-shan:
Refer to the metadata manager and design to ensure that the attributes shown in the UI have the exact type.

PR #177, opened 2024-07-02
#176: add other meta fields to solrObject type

pengyin-shan:
This PR addresses #176. It is a simple update to the Solr object type before moving on to the next stage of UI development. By adding clear types for attributes, we can avoid many "if" loops in the data processing.

I'll continue to update the attributes list while working on #172.


mradamcox (approved):
Looks good, thanks!

PR #178, opened 2024-07-09
Implement url params

mradamcox:
This PR includes a few examples of how we can use url params to drive search and page state, and the general process is described below. Parts 1 and 2 could be split across different components, or exist together on the same component.

#### Part 1: Push values to params

On the component that has an input element, `updateSearchParams()` (which is currently located in `ManageURLParams.tsx`) must be called somewhere to push values from input elements into the proper URL param. To use this function, path and router information must be collected at the top of the component. Then, for any checkbox or text input, create something like `handleInput` which will sanitize the input (if needed) and then call `updateSearchParams`. A simple example looks like this, where a dropdown selection gets pushed into the `?query=` param:

```
import { useRouter, usePathname, useSearchParams } from "next/navigation";

etc.
 {
    const searchParams = useSearchParams();
    const currentPath = usePathname();
    const router = useRouter();

    const handleDropdownSelect = (event, value) => {
        updateSearchParams(router, searchParams, currentPath, "query", value, "overwrite");
    };
}
```

There are better examples in the PR commits here.

updateSearchParams() alters URL params without navigating to a new page or adding an entry into the browser history. There is a bit of extra logic in the function to handle multi-valued entries (these are separated by |, as in `?layers=state|county`), and allows you to specify whether a param value should be added, removed, or overwritten (the first two being necessary for multi-values, while the latter should be used for basic params like `?query=social+vulnerable`.

#### Part 2: Using values from params

URL params can directly drive behavior on the page if hooks have been set up to watch them like this:

1. `const searchParams = useSearchParams();` at the top of any component will create a reactive variable that changes whenever params are changed
3. A `useState` variable is made for each important category of URL params, like `query` in this case, or `layers` which is used to determine which layers will be shown on the map.
4. A "high-level" `useEffect` hook is constructed that watches the url params, and when they change, updates any performs the search that updates the individual state variables that have been created above.
5. One or more "low-level" `useEffect` hooks are created for each of the state variables, and they are used to manipulate the page as needed. Manipulations could be:
    1. Updating the results list by running a new query to solr using all filters.
    2. Showing/hiding the item detail section
    3. Updating the map by add/remove layers or overlays, or zooming to specific locations.

```
  // create queryString variable as state
  const [queryString, setQueryString] = useState(null);

  // create variable to get search params
  const searchParams = useSearchParams();

  // whenever search params are updated (i.e. through updateSearchParams) this hook
  // will run, and if the param in question, "query", has been changed, setQueryString
  // will update the queryString state
  useEffect(() => {
    const newQueryString = searchParams.get("query");
    if (newQueryString && newQueryString != queryString) {
      setQueryString(newQueryString);
    }
  }, [searchParams]);

  // whenever the queryString state changes, do something on the page. In this case, run search
  useEffect(() => {
    if (queryString) {
      searchQueryBuilder.suggestQuery(queryString);
      handleSearch(queryString);
    }
  }, [queryString]);
```

Overall, this works, but I'm getting a lot of warnings while running `yarn build` like this:

```
React Hook useEffect has a missing dependency: 'queryString'. Either include it or remove the dependency array
```

To address this I seem to be able to add `queryString` as a dependency of the first `useEffect` above without any side effects, but it seems circular to do that, given that `queryString` is altered within the hook by calling `setQueryString`.

**Version update**

I had to upgrade next to a later minor release to properly handle the typing in router.replace(), so make sure to remove node_modules and re-run `yarn install` to test this out.


pengyin-shan (approved):
Looks good to me. We should also prepare a view (or a separate page) to show what will be displayed if the user enters a wrong URL under the search domain.

mradamcox:
Hm, could describe what you mean by wrong URL?

pengyin-shan:
> Hm, could describe what you mean by wrong URL?

Ah, sorry, I meant if a user copies a URL but pastes it incorrectly (for example, a result list view with a URL including an item ID, but when pasting the url somewhere else, a character is lost), then we could have a '404' or 'item not found' page prepared. Otherwise, there could be awkward errors.

mradamcox:
Ok, I'm thinking you mean something like `?show=incorrect-id`? Yes, in this case the item details component would just show "item not found" or something, and closing that component would remove the `show` param altogether, as normal.

mradamcox:
Ok, I've fixed the `React Hook useEffect has a missing dependency: 'queryString'. Either include it or remove the dependency array` errors that I had introduced, though there are still some more caused by other `useEffect` hooks that were in the app before. Ready to merge now.

PR #179, opened 2024-07-12
#173: result item details panel

pengyin-shan:
This PR is for #173, matching the design of the item detail panel:
Screenshot 2024-07-12 at 12 21 53 PM.

In this PR, I have also:
1. Made the `useRouter` imported from 'next/router' instead of 'next/investigation' based on https://nextjs.org/docs/pages/api-reference/functions/use-router.
2. Refactored the result item to accept a SolrObject as a parameter instead of a list of single variables, making the code cleaner.
3. Added the link between the result item, item detail, and the original map view.

Note that the `IconMatch` method now only includes the dataDiscoveryIcon as an example. Once we finish defining the matching rule between attributes (I assume it would be theme or resource_type) and icons, we can complete this method to display the correct icon based on the attribute.

## How to Test

1. Navigate to http://localhost:3000/search
2. You should see the map view since there's no show parameter added in the URL.
Screenshot 2024-07-12 at 12 29 34 PM

3. You should also see three result items on the left as examples.
4. Click one of the 'View' links for the result item. You should see the map view replaced by the detail panel view:
Screenshot 2024-07-12 at 12 29 19 PM

5. The detail panel view should include all attributes listed in the design. I believe I used the correct URL for the link, but please let me know if I misunderstood this.

6. On the top right of the detail panel, click the 'Share' button. You should see a Share this URL modal appear, allowing you to copy the current URL and close the modal:
Screenshot 2024-07-12 at 12 31 27 PM

7. Click the 'copy' icon, and you should see a pop-up to confirm the copy action (we can update this behavior later if we want). Now, paste this URL into another browser window. You should reach the item detail view page directly instead of seeing the map.
Screenshot 2024-07-12 at 12 42 10 PM

8. Click the 'Access' button at the top right of the detail panel. You should see a new tab/window opening, leading the user to where the dataset is:
Screenshot 2024-07-12 at 12 34 06 PM

9. Click the 'Back to map view' on the top left of the detail panel. You should see the map view return, and the URL should be cleared.

I'll continue to work on the top search bar after this component.


mradamcox (changes requested):
Looks good overall, however, if I search the from the main search box (type "social" and hit enter, for example) I'm now getting a `TypeError: props.resultItem is undefined`. Can you look into that before we merge this?

Also, note that there are some issues with the geography map layers that came after some work on those source files yesterday, so that'll be fixed next week.

pengyin-shan:
Thanks for finding this, @mradamcox. I didn't link the result list to the search bar before, which caused this issue. I have updated the code to link the search bar with the new result card and added null value handling for some fields. 

1. Now, when you go to the search page, you should see all result items matching all datasets we have in Solr:

Screenshot 2024-07-12 at 5 29 07 PM


2. You need to scroll down to the search bar on the left side. Type any word and click 'search'; you should see the corresponding results:

Screenshot 2024-07-12 at 5 29 41 PM


3. Click 'View' for any result item, and you should see the corresponding detail panel. After we get the new search component and add the page layout, this part will be modified to fit the new situation.

mradamcox (approved):
Great, works now. Thanks!

Issue #181, opened 2024-07-17
Render HTML in Description and Display Notes


mradamcox:
In the full item detail page, we should render the description and various notes fields as HTML. This would allow us to add links within the description metadata fields, which has been requested by folks working on the metadata records (it would really enrich the entries, as well).

pengyin-shan:
Do we have a result item with description and various_notes fields that I can use to test this? 

And can you explain render as HTML?

mradamcox:
Sure, I have added this phrase to the Description field of the Social Vulnerability Index on the dev core: `This is an <a href="https://sdohplace.org">example link</a> that should be rendered properly on the front-end.`. When that field is displayed in the item details section, it should look like:

> This is an [example link](https://sdohplace.org) that should be rendered properly on the front-end.

pengyin-shan:
Ah I see, thanks for your explaination!

mradamcox:
Ok, I added a link in the Neighborhood Atlas entry to test, and looks like we'll need more work on this, as the description currently gets cut off at the beginning of the HTML tag: https://discovery-app--cheerful-treacle-913a24.netlify.app/search?show=neighborhood-atlas

pengyin-shan:
> Ok, I added a link in the Neighborhood Atlas entry to test, and looks like we'll need more work on this, as the description currently gets cut off at the beginning of the HTML tag: https://discovery-app--cheerful-treacle-913a24.netlify.app/search?show=neighborhood-atlas

I see, I'll check on what happened.

pengyin-shan:
I tested random hyperlinks and I could not replicate the issue locally...may I know the full description text we used for https://discovery-app--cheerful-treacle-913a24.netlify.app/search?show=neighborhood-atlas?

pengyin-shan:
@mradamcox for this issue, do we still have the item with links in description that I can use for test?

mradamcox:
Sorry yeah, just looking at this. The discovery_app branch deploy seems to be broken now somehow, I guess it must have something to do with the index. Looking into this now...

mradamcox:
The problem was [this line](https://github.com/healthyregions/SDOHPlace/blob/b8ef9fa58e4e76125167097d258118bceda46432/meta/helper/solrObjects.ts#L21), the description was being truncated to its first sentence, which cut the link in half. There is no need to truncate the description anyway though. Technically it is an array, but passing this array into DOMPurify.sanitize() works just fine.

PR #182, opened 2024-07-17
#181: add html parsing to description, usage tips and display_notes

pengyin-shan:
This PR is for #181. I changed the description, usage tips, and display notes for the HTML rendering style.

Currently, I haven't found any result items to test, but you can modify the code to replace the props value with some HTML code, and the result should be rendered in HTML style.

Please let me know if we want other fields to follow this pattern.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "axios": "^1.6.8",
     "classnames": "2.3.2",
     "date-fns": "^2.30.0",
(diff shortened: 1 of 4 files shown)


mradamcox (approved):
Perfect, thank you for using DOMPurify! I was just reading more about this and was going to ask that we use it :). These fields will be good for now, and the pattern should be easy to follow for others down the road.

Issue #180, opened 2024-07-15
Create component for the search area of the new discovery platform


pengyin-shan:
Follow the design

PR #183, opened 2024-07-19
#180: create the search component based on the new UI design

pengyin-shan:
This PR is for #180. It creates the search panel at the top of the search page based on the new design. This PR still uses the previous methods of handling parameters. After creating the frame page to include all components, the method will be reviewed and updated. Some width ratios in the search area may also be adjusted after including other components on the page.

## How to Test

1. Navigate to the search page. The search area is placed at the top of the screen.

2. For spatial resolution checkboxes, I used the original list of spatial resolutions. Among them, the State and County checkboxes are set to be checked by default and uneditable (we can remove them as well). Other checkboxes can be turned on and off. The map will reflect the changes.

Screenshot 2024-07-19 at 11 56 40 AM

3. The search input field has been modified to reflect the new UI design. The 'Advanced' button is currently not functional; I placed it there as a placeholder.

4. Begin to type and search. The autocomplete dropdown appears when the user begins to type, and the user can click the pre-populated data or the search button to see the result, as before.

5. After submitting a query and getting the result, copy and paste the current URL into another browser. The state of the page should be maintained in the new window.

### DIFF
--- a/src/components/map/mapArea.tsx
+++ b/src/components/map/mapArea.tsx
@@ -144,7 +144,7 @@ export default function MapArea({
       const map = mapRef.current.getMap();
 
       let selectDisplayLayers = new Array<LayerDef>();
-      // get all needed interactive layers based on current presented results's spacial resolution
+      // get all needed interactive layers based on current presented results's spatial resolution
       searchResult.forEach((result) => {
         if (result.meta["spatial_resolution"]) {
           const spatial_res =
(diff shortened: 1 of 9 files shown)


mradamcox (changes requested):
Coming along very well. Just a couple of things, both should be fairly simple:

1. Ensure the correct font-family is in the input (the dropdown is correct but fix the placeholder and input text)
2. Change how State and County checkboxes are treated so they are the same as all other boxes (not checked by default)

I know we have had the State and County boxes checked by default in the past, but that was more a workaround to show those layers on the map by default. I'd like now to change this because it's more important that those checkboxes will function as normal filters on the results list just like the others.

Checking each box should:
1. update the `?layers=` param, which in turn drives map layer addition/removal
2. update the spatial_resolution search filter, which will be in a URL param but we'll cross that bridge when we get into all the other filters.

mradamcox (approved):
Looks good now, thanks!

PR #214, opened 2024-07-26
#180: create new search page

pengyin-shan:
This PR is for #180. I renamed the old search index page to `index_back.tsx` and created a new index page that combines the new result page components to match the updated UI design. This PR mainly targets UI adjustments, as these changes are substantial enough to warrant their own PR.

I also fixed the input field error in this PR, which was observed in last week's demo. Since the search box has been extracted as a distinct component, some input controls needed adjustment to coordinate with this change. This PR includes that fix. Further adjustments to the input fields may be made when working on the new filter functions.

## How to Test

1. Navigate to the search page. You should see that the layout matches the design.

-----

2. The result list panel should show "All data sources" without a "Related" field since the user hasn't done any search. Note that the process bar is controlled by the browser, so it appears when the user hovers over the result list. The user should also notice that the result list doesn't end until the bottom of the page:

Screenshot 2024-07-29 at 2 39 19 PM
Screenshot 2024-07-29 at 2 40 33 PM

The map side should have headers and the bottom description part added. The two buttons on the bottom of the map side should link to the correct places:

Screenshot 2024-07-29 at 2 44 38 PM

-----

3. Search for anything in the input field. You should see the result panel change to "Results" as the header, and a "Related" field should be added (currently, the result list is used since we haven't decided what should be categorized as "related"):

Screenshot 2024-07-29 at 2 47 30 PM

Note that if your search results contain more than three items (e.g., "Wisconsin" or just an "a"), the result list will be scrollable, and the height of the map will match the height of the result list:

Screenshot 2024-07-29 at 3 53 16 PM

-----

4. Gradually press the 'delete' key to remove all characters in the input field, or select all characters in the input field and delete them at once. You should notice that once all characters are removed, the page resets to step 1 (but doesn't affect the spatial resolution status):

Screenshot 2024-07-29 at 2 48 37 PM

-----

5. Click any "View" for a result list item. You should notice the corresponding item gets highlighted, and you can review the details on the right panel:

Screenshot 2024-07-29 at 2 50 28 PM


pengyin-shan:
Noticed a map rendering issue. Working on the fix of it.

pengyin-shan:
Fixed error. This PR is ready for review.

mradamcox (changes requested):
Looking really good, though there are a few design/spacing elements to improve, looking back at the mockups.
1) Within the card for each search result item, reduce spacing between lines to match mockup
2) Reduce space above the header bar. Looking at the mockups, there is about equal space above the navbar text as there is below it. I realize that moving the header bar up will overlap the dots and lines, but I think we need to fix those separately--they are currently stretched and spaced much differently than they should be.
3) Per changes in the mockups, swap the search bar and filter set with the header and explanatory text. To me, this raises other questions about the width of that entire section, which we will discuss later today.

pengyin-shan:
> Looking really good, though there are a few design/spacing elements to improve, looking back at the mockups.
> 
> 1. Within the card for each search result item, reduce spacing between lines to match mockup
> 2. Reduce space above the header bar. Looking at the mockups, there is about equal space above the navbar text as there is below it. I realize that moving the header bar up will overlap the dots and lines, but I think we need to fix those separately--they are currently stretched and spaced much differently than they should be.
> 3. Per changes in the mockups, swap the search bar and filter set with the header and explanatory text. To me, this raises other questions about the width of that entire section, which we will discuss later today.

Interesting, it seems there are some new UI changes that were made just after I finished this PR. The original design had the header positioned between two dots, with the search bar and filter set on the left side of the page. I'll hold off on making any changes until I confirm that these are the final decisions, in case new changes come today and I can combine them to update this PR.

pengyin-shan:
@MrAdamCox : I updated the result card text gap and the search bar row to match what we discussed yesterday. 

I switched the positions of the search box and title text (Data Discovery) and reduced the top and bottom margins. I also replaced the background lines with the two lines from the search mockup and adjusted their positions. 

Please let me know if these changes look better to you. Thanks!

mradamcox:
Looking great! I will merge now. The dotted lines still shift around a bit as the page resizes, in the same way that the lines on the other pages do. However, we can tackle that later, as well as trying out max-width on a central container. Ready to merge, but because this will override the old filters that are on the bottom of the discovery_app build, I'll wait to confirm those are no longer needed for design purposes.

pengyin-shan:
> Looking great! I will merge now. The dotted lines still shift around a bit as the page resizes, in the same way that the lines on the other pages do. However, we can tackle that later, as well as trying out max-width on a central container. Ready to merge, but because this will override the old filters that are on the bottom of the discovery_app build, I'll wait to confirm those are no longer needed for design purposes.

Sure, if we want to keep this until the filter design is finished, it should be OK. 

On the other hand, I did make an `index_backup.tsx` file that includes all the old filters and previous components. If we want, I can use this as a backup page for reference, and we can merge this PR then. Either way will work.

mradamcox:
Ready to merge now that design is completed for filters panel.

PR #260, opened 2024-08-16
Fix the infinitely re-rendering of DiscoverArea and its subcomponents

pengyin-shan:
This PR addresses the issue of infinite re-rendering in `DiscoverArea` and its subcomponents, which can be observed by adding a console log in the component. The root cause was the `useEffect` in `index.tsx`, which is responsible for providing `solrObjectResults` to all components in `DiscoverArea`. Without dependencies, the `useEffect` was running on every render, causing `setSolrObjectResults` and other state setters to trigger additional renders, leading to the infinite loop.

I've added solrUrl and schema as dependencies, ensuring that `<DiscoverArea>` and its subcomponents only re-render when these two variables change. I also cleaned up some unused code in `index.tsx` after making this change.

## How to Test
Run the app and navigate to the search page. You should observe only two render cycles, as indicated by the console message in the `ResultPanel` component. The two renders are expected in development; in production, it should render only once.
Screenshot 2024-08-16 at 12 32 36 PM

Please let me know if this makes sense to you. Please feel free to add more test messages to confirm the fix.

### DIFF
--- a/src/components/search/resultsPanel/resultsPanel.tsx
+++ b/src/components/search/resultsPanel/resultsPanel.tsx
@@ -24,7 +24,7 @@ const useStyles = makeStyles((theme) => ({
 
 const ResultsPanel = (props: Props): JSX.Element => {
   const classes = useStyles();
-  
+  console.log("ResultsPanel", props.resultsList);
   return (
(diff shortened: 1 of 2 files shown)


mradamcox (approved):
Perfect, and always good to clean up some extra stuff too :)

PR #262, opened 2024-08-16
Update map component nuqs

mradamcox:
Some updates to the map component, still more work to do but it's a start.

- Replace layers param handling with `nuqs` per #235
- Add a `bbox` url param to track map panning, and use this param for initial load of map (if present)
- Style zoom to region buttons per mockups
- Add a "Search this area" checkbox and corresponding url param `bboxSearch=true`

@pengyin-shan For future reference, the idea is that if `bboxSearch=true`, then an additional filter will be applied to the search query that casts the value of the `bbox` param (i.e. the current extent of the map view) into a Solr-friendly query and runs against one of the geometry-related fields in the index, like `geometry` or `bounding_box`. In other words, every time the map is moved, the search results will update based on the bbox param, if that box is checked. We don't have metadata ready to really implement this yet, so I see it as a task that I'll come back to a couple weeks from now. Also, this feature isn't in the mockups, it's just something that I wanted to stub out because I have a feeling it will ultimately be desirable.

### DIFF
--- a/src/components/map/mapArea.tsx
+++ b/src/components/map/mapArea.tsx
@@ -36,6 +36,8 @@ import {
 } from "../../components/map/helper/layers";
 import { sources } from "../../components/map/helper/sources";
 
+import { ZoomButton, EnableBboxSearchButton } from "./mapButtons";
+
 import getCountyGeo from "./helper/countyLocation";
 import CheckBoxObject from "../search/interface/CheckboxObject";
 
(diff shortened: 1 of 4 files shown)


pengyin-shan (commented):
Hi @mradamcox,

I just quickly skimmed through your approach, and it looks good. I haven't tested it yet, but we can do that once the data is ready as you said.

One thing for discussion: Yesterday, I noticed that we might end up with many parameters in the URL and multiple useQueryState calls, which could lead to unexpected results since we might lose track of them over time. To address this, I created `ParameterList.tsx` to centralize all possible parameters. This way, we can handle all useQueryState calls in one place. Then, `DiscoveryArea`, as the parent component, can call this once and retrieve all of the query states, like:
Screenshot 2024-08-16 at 5 01 55 PM
then
Screenshot 2024-08-16 at 5 02 18 PM
How do you feel about this approach v.s. putting `useQueryState` in subcomponents? This method will help us to always know where `useState` is called and control which parameters are in the URL. 

This is just a side discussion as we move towards adding parameters. I look forward to hearing your thoughts!

mradamcox:
@pengyin-shan I did see ParameterList and it seems like a good approach to try out, once we have solidified all of the possible parameters. Just for now I think using `useQueryState` in separate components is fine, because nuqs requires so little overhead to get working. But, ultimately a "registry" like this would be wise to implement.

pengyin-shan:
> @pengyin-shan I did see ParameterList and it seems like a good approach to try out, once we have solidified all of the possible parameters. Just for now I think using `useQueryState` in separate components is fine, because nuqs requires so little overhead to get working. But, ultimately a "registry" like this would be wise to implement.

Okay good. I just want to make sure we're on the same page about handling this, since we'll later need to combine filter and sort options with bbox (or other map query parameters) to get the updated result list. 👍

pengyin-shan (approved):

mradamcox:
For sure. Will merge this PR in now.

PR #266, opened 2024-08-19
[WIP] Issue 234 filter UI sort with api

pengyin-shan:
Draft PR for sorting and filters


mradamcox (approved):
Great work. Thanks for working to set up search requests directly to Solr, this is really good and now scalable.

The work raises the need for a few more buttons, like a general "clear filters" button, etc., but we'll get to that after this is implemented. Thanks for this great work!

Issue #212, opened 2024-07-24
[TASK] Make the share link to be in page (following new UI)


pengyin-shan:

mradamcox:
Completed.

PR #267, opened 2024-08-22
#181 ,#212: Created the new UI for share link and fix the html rendering issue on description

pengyin-shan:
This PR addresses issues #212 and #181. It introduces a new UI for the "Shared" function and generates a placeholder link if the description is not yet complete.

## How to Test
1. Go to the search page and select "Neighborhood Atlas."

2. You should see that the description has been replaced with "the Link." This is because the data source for this description is not yet complete in Solr, so the HTML cannot be rendered like other result items. I have added temporary code in `headerRow.tsx` to create this placeholder link, which can be removed once the data is complete:
Screenshot 2024-08-26 at 5 15 56 PM

3. Click the "Share" button, and you should see the new UI style:
Screenshot 2024-08-26 at 5 16 17 PM

4. Click the "Copy" icon in the "Shared" panel. You should see a notification indicating that the link has been copied.

5. Paste the link to a new window. You should see the same page appears.


mradamcox (approved):
Looks good to me. I'm surprised the link in the HTML is so problematic, as the record looks straightforward in the Solr response: 
image
I can take a look at it soon too. Thanks for all this work!

Issue #257, opened 2024-08-14
Implement glossary-style help/tips within Data Discovery app interface


mradamcox:
We should implement some sort of help/tooltips that will define certain terms within the search interface. Ideally this could be shown for terms not only in text paragraphs, but also search filter values, etc.

mradamcox:
Need to remove the orange border around the tooltip panel but then this ticket can be closed.

PR #270, opened 2024-08-27
feat: create popover for glossary items

mradamcox:
Initial implementation of #257, a "glossary" that can power extra definitions for key terms. It is implemented with a simple MUI popover that has a style matching the buttons within the map interface. The popover appears on click (not on hover) so it should work fine on mobile as well. 

image

image

By default, it looks for a predefined entry in the `meta/glossary.json` file, but any content can be passed to it, if there are context-specific definitions we want to display.

To test:

1. go to: https://deploy-preview-270--cheerful-treacle-913a24.netlify.app/search
2. click on the underlined term "health equity" in the text description

We could implement this anywhere within the website, besides news posts and showcase pages (though, it would be possible to set that up). For button labels, we could use this same style but make the popover [appear on hover](https://mui.com/material-ui/react-popover/#mouse-hover-interaction) (so as not to interfere with existing click events).

### DIFF
--- a/meta/glossary.json
+++ b/meta/glossary.json
@@ -1,0 +1,3 @@
+{
+    "health equity": "short description of health equity"
+}
--- a/src/components/GlossaryPopover.tsx
+++ b/src/components/GlossaryPopover.tsx
@@ -22,7 +22,7 @@ export default function GlossaryPopover({ entry }) {
     return (
       <>
         <button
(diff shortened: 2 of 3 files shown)


mradamcox:
@shubhamk008 I'm going to go ahead and merge this, but we can easily make changes down the road if needed. Seems like the core concept here is good enough to go ahead with.

Issue #273, opened 2024-08-28
Clicking spatial resolution checkboxes should update search results


mradamcox:
Checking these boxes should directly update a `spatial_resolution` param in the URL, and trigger filtered search results, just like if a box was checked in the filters panel (this is the pattern we have been preparing for with `nuqs`).

This should be an OR filter--if two boxes are checked then any record with a spatial_resolution value matching either box should be returned.

PR #282, opened 2024-08-28
#273: Add filter function for spatial resolutions and create the new UI

pengyin-shan:
This PR addresses #273. It includes adding the filter function to the spatial resolution checkboxes and updating the UI to reflect the new design for this section. I also updated a few filter methods to remove redundancy (this will be an ongoing process to optimize the code).

## How to Test
1. Navigate to the search page. No spatial resolution filter should be selected.

2. Select one of the spatial resolution checkboxes. The results list should be updated based on the selection, and the map layer should update accordingly:
Screenshot 2024-08-28 at 6 08 41 PM

3. Select multiple spatial resolution checkboxes. The results list should be filtered by any of the selected spatial resolutions, while all relevant layers should be added to the map.
Screenshot 2024-08-28 at 6 08 56 PM

4. Uncheck the checkboxes you selected in steps 2 and 3. The layers and results list should adjust accordingly.

5. Repeat steps 2 to 4 after applying other filters and queries.
Screenshot 2024-08-28 at 6 09 19 PM

### DIFF
--- a/meta/_metadata/layer_match.json
+++ b/meta/_metadata/layer_match.json
@@ -1,9 +1,9 @@
 {
-  "City": "Place",
-  "Zip Code Tabulation Area (ZCTA)": "Zip Code Tabulation Area (ZCTA)",
-  "County": "County",
-  "State": "State",
-  "Census Tract": "Census Tract",
+  "City": "place",
+  "Zip Code Tabulation Area (ZCTA)": "zcta",
+  "County": "county",
+  "State": "state",
+  "Census Tract": "tract",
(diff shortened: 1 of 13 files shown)


mradamcox (approved):
Thanks, this looks good. Per the ticket, I did want a new param, `spatial_resolution` to be added instead of using the existing `layers` param, but that should be easy to modify after I merge.

mradamcox:
Honestly, we may want these to be connected afterall, based on some of the suggested filter-by-zoom-level behavior we discussed the other day.

PR #286, opened 2024-09-04
#238: Update the search box to match the new design

pengyin-shan:
This PR is for #238. It updates the search box to match the new design and adds a "close" button that allows users to reset the search while typing. Note that the auto-complete is not working because `/suggest?` does not return any results at the moment. The Info icon is currently kept as a "/#" link until the corresponding page is designed.

## How to Test
1. Navigate to the search page and type anything in the search box.

2. You should notice an "X" button appears:
Screenshot 2024-09-04 at 9 58 03 AM

3. Click this button, and you should see all search input cleared.

4. Try adding a few spatial resolution filters and then perform a search.
Screenshot 2024-09-04 at 9 58 42 AM

5. Now click the "X" button again. You should see that the filters are maintained, but the result list is refreshed by removing the query.
Screenshot 2024-09-04 at 9 58 55 AM

6. Enter something in the search box, perform the search, and then paste the URL into another window. You should see the close button is still present in the search box in the new window.

### DIFF
--- a/src/components/search/searchArea/searchBox.tsx
+++ b/src/components/search/searchArea/searchBox.tsx
@@ -87,7 +87,9 @@ const CustomPaper = (props) => {
 };
 
 const SearchBox = (props: Props): JSX.Element => {
-  const [showClearButton, setShowClearButton] = React.useState(false);
+  const [showClearButton, setShowClearButton] = React.useState(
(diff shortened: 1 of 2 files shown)


mradamcox (approved):
Looks great, thanks!

Issue #264, opened 2024-08-19
"No results" on discovery app needs to be adjust to fix the new UI


pengyin-shan:
Currently the old "No results" view is still being used. Need to adjust this to fit the new UI

mradamcox:
![image](https://github.com/user-attachments/assets/a660beaf-06fe-4e9a-a7b9-881fe0bd08f6)

We can put placeholder tags in for the theme searches for now, we will connect them once we have that field ready to be filtered.

PR #289, opened 2024-09-05
#264: make the "No results" match the new design

pengyin-shan:
This PR addresses #264. It updates the "No results" section to reflect the new design, along with some style adjustments to the theme icons. Additionally, I updated the .gitignore file to include `nvmrc` for Node version control, as we're now using Node 18.

## How to Test
Run the app and search for a term (e.g., "a"). You should see the "No results" message appear with hard-coded theme suggestions. You can click the "X" in the search box to clear the search at any time:

Screenshot 2024-09-05 at 1 35 04 PM


mradamcox (changes requested):
This looks great, but I think we can actually hookup the "theme" buttons to apply actual search filters. Note that clicking one of these buttons must 1) clear all existing queries and filters, and 2) add a `subject` search filter based on the name of the "theme".

pengyin-shan:
Link bug #291 to this PR, as it needs to be addressed before the subject filter will work.

pengyin-shan:
@mradamcox, I’ve updated the code to make the subject icons functional. After adding this functionality, we’ll need to incorporate a design element (such as a button) on the page to reset everything to its initial state. I’ve outlined the steps below. Additionally, I fixed the spatial resolution checkbox behavior (#291) and added `spatial resolution` as a separate URL attribute, instead of using only the `layers` parameter as we did previously.

## How to Test
1. Run the app and navigate to the search page.

2. Select "State" from the spatial resolution options. You should now see the map displaying the new state layer:
Screenshot 2024-09-07 at 12 31 33 PM

3. Drag the right pin of the "year" slider to 1976. You should see "No results" because there is no data for spatial_resolution=state and index_year<=1976:
Screenshot 2024-09-07 at 12 33 14 PM

4. Click the "Community Health" icon. This icon has been modified so the subject filter returns data for testing. Once the database is updated, we can change it back to "Transportation."

5. You should now see all filters reset, and the data filtered by subject=Community+Health. The URL should also update accordingly:
Screenshot 2024-09-07 at 12 34 21 PM

6. @shubhamk008: We’ll need a "reset" button somewhere after the subject icon is clicked to trigger an automatic filter action. Currently, we only have the "X" in the search box that performs this action, and it only appears after the user enters something in the search box.

7. Type "a" in the search box, then click the "X" that appears. The page should reset to its initial state (i.e., all filters cleared, including the subject filter):
Screenshot 2024-09-07 at 12 37 32 PM

8. Select a spatial resolution, then enter "a" in the search box and search for it. The map layers should remain visible, but no result view will appear. The URL should include both spatial_resolution and layers:
Screenshot 2024-09-07 at 12 38 30 PM

9. Repeat step 4. You should see that all filters (including spatial resolution) are cleared, except for the "Community Health" subject filter:
Screenshot 2024-09-07 at 12 40 21 PM

10. Copy the URL and paste it into another window. You should see the same view in the new window.

The overall behavior of the filters is still not stable, so if you find any issues, please post them, and I’ll address them after I return from my trip. Once we add more functions, we can refine the filters again to ensure everything works correctly.

mradamcox (approved):
Great, looks good. The `spatial_resolution` param is what I was thinking too. Thanks!

Issue #265, opened 2024-08-19
Review `handleSearch` to adjust suggestionBuilder once suggestion endpoint is fixed


pengyin-shan:
HandleSearch should be updated to accommodate the new approach, and the suggestionQueryBuilder may need to be removed after HandleSearch is widely adopted for conducting searches exclusively. Review this method after the suggestion endpoint is fixed.

pengyin-shan:
Hi @mradamcox, no rush, but do you have any idea why our suggestion endpoint isn't working? For example, https://solr.sdohplace.org/solr/blacklight-core-dev/suggest?q=education isn’t returning any suggestions. I’m wondering if a re-ingestion might help 🤔

mradamcox:
You are right, it needed to be fully reindexed (all records deleted and then recreated from our JSON metadata records). I assume it stopped working because I had tried some alterations to the schema for the suggestion field, but didn't realized I had to fully reindex to have them be implemented. Thanks for the suggestion!

pengyin-shan:
Great, thank you! Once the `suggest` endpoint is working again, I will refine the updated search methods and fix the autocomplete feature for the search box.

mradamcox:
Ok all set, it is now working on both the Solr cores.

pengyin-shan:
awesome, thanks!

PR #299, opened 2024-09-19
#265: fix the auto suggestor

pengyin-shan:
This PR is for #265. It enables the auto-suggest feature after the user inputs anything in the search box.

## How to Test
1. Go to the search page and type anything in the search box. You should see the auto-suggestion feature enabled again:
Screenshot 2024-09-19 at 11 06 05 AM

2. Click on one of the results (for example, southwest coastal alaska). You should see the result list appear after the query:
Screenshot 2024-09-19 at 11 08 14 AM

3. Run a sort or filter, and the result list should adjust as expected:
Screenshot 2024-09-19 at 11 08 24 AM

## NOTE
@mradamcox : while fixing the suggestion feature, I noticed an issue that I want to confirm with you. One example is "southeast coastal Alaska," which was suggested by Solr. Our existing rule for generating the parent object from the child object resulted in no results.

When I checked the query (https://solr.sdohplace.org/solr/blacklight-core-dev/select?q=southeast%20coastal%20alaska&fq=&rows=1000), it appears that all the results have a parent of "tree-canopy-cover." However, the returned result doesn't include a "tree-canopy-cover" item. Instead, we only have items like "tree-canopy-cover-alaska2013" and "tree-canopy-cover-alaska2014."

On the other hand, we do have a "Tree Canopy Cover" object in the database. If a user enters "Tree Canopy Cover" directly and then selects any of the returned results from the drop-down or directly searches this term, it will appear. However, this item doesn't include "southeast coastal Alaska," so it doesn't show up in the above search.

To address this, I changed the rule so that if the parent item is not in the returned list, all the child objects are shown instead. Please let me know if this change makes sense. We can also update the Solr rule to ensure that if all child items are returned, the parent item must also be included.

### DIFF
--- a/meta/helper/solrObjects.ts
+++ b/meta/helper/solrObjects.ts
@@ -1,3 +1,4 @@
+import { c } from "nuqs/dist/serializer-C_l8WgvO";
 import { SolrObject } from "../interface/SolrObject";
 // import { SolrParent } from "../interface/SolrParent";
 import { findFirstSentence, schemaMatch } from "./util";
@@ -71,6 +72,8 @@ const generateSolrParentList = (
     .filter((solrObject) => solrObject.parents)
     .forEach((childObject) => {
       childObject.parents.forEach((parentTitle) => {
+        // only if the object has parents, add the parent to the result
+        if(solrObjects.filter((solrParent) => parentTitle === solrParent.id).length >0 ) {
         solrObjects
           .filter((solrParent) => parentTitle === solrParent.id)
           .forEach((solrParent) => {
@@ -83,6 +86,10 @@ const generateSolrParentList = (
               : null;
             result.add(solrParent);
           });
+        }
+        else {
+          result.add(childObject);
+        }
       });
     });
   solrObjects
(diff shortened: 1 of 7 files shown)


meta/helper/solrObjects.ts
```diff
@@ -1,3 +1,4 @@
+import { c } from "nuqs/dist/serializer-C_l8WgvO";
```
mradamcox, line 1: Was this an intentional import? I don't see it used and maybe it was auto-added by your IDE?

src/components/search/discoveryArea.tsx
```diff
@@ -17,6 +17,7 @@ import {
   updateAll,
 } from "./helper/ParameterList";
 import FilterPanel, { grouped } from "./filterPanel/filterPanel";
+import { c } from "nuqs/dist/serializer-C_l8WgvO";
```
mradamcox, line 20: See above comment

mradamcox (changes requested):
Looks good, and the exception you pointed out is noted. We'll think about the parent/child relationships more soon, too. Could you just check these nuqs imports I pointed out, and then ready to merge.

src/components/search/discoveryArea.tsx
pengyin-shan: Thanks for catching that! Those were auto-imports, and I've removed them. As for the commit related to the reset button, you can view it in [PR #301](https://github.com/healthyregions/SDOHPlace/pull/301) instead.

mradamcox (approved):
Great, thanks!

Issue #287, opened 2024-09-04
Match icons with themes


mradamcox:
We need to match the new icons @shubhamk008 has selected with the 12 themes that will be implemented as a controlled vocabulary for that field.

pengyin-shan:
The themes will be added as a new attribute to the Solr object, right?

mradamcox:
No, there is an existing field `subject` which will hold what we are calling "themes" here (this is a little confusing because there is a different field `theme`, but that one has a pre-defined controlled vocabulary, so we are using `subject` which allows us to create our own controlled vocabulary).

Thing is, we still need to fix the existing metadata records so that the `subject` field properly reflects the options listed above.

mradamcox:
I think the best way to implement this on the frontend for now would just be a lookup table that matches these values with icons, and ignores any `subject` values that don't match the pre-made lookup. That way we aren't dependent on fixing all the metadata, but once it does it fixed the system will display it properly.

pengyin-shan:
Currently, I have an IconMatch method to generate icons based on certain strings:
<img width="575" alt="Screenshot 2024-09-05 at 11 28 20 AM" src="https://github.com/user-attachments/assets/50bbac79-de7d-445c-a5b6-177a7ec950f3">

Once we are done with the subject values in Solr, these values can be added to this method and matched with the icons Shubhma provided.

mradamcox:
Ok, great, that will work for sure.

PR #304, opened 2024-09-20
Add matched theme icon and enable all subject filters

pengyin-shan:
This PR addresses #287, #296, and #297. It adds matching MUI icons to subjects in the result list and detail panel and enables the filter operation when someone clicks on them.

Additionally, this PR includes a fix for #305. When a user opens the detailed view and then applies a filter that triggers the "no results" view, the map view will be restored instead of leaving the page blank.

## How to Test
1. Navigate to the search page. You should see that all result item header images have been updated to match their corresponding subject:
Screenshot 2024-09-20 at 5 03 37 PM

2. Click on an item with the "Compose" icon. You should see more than one subject tag for this item in the detailed view, with the icon in the title also updated:
Screenshot 2024-09-20 at 5 07 11 PM

3. Click on an item with another subject icon. You should see the corresponding subject tag for this item in the detailed view:
Screenshot 2024-09-20 at 5 08 12 PM

4. Click the subject icon in the detailed view. The result list should now be filtered by this subject:
Screenshot 2024-09-20 at 5 08 34 PM

5. While in the detailed view, apply a spatial resolution filter that does not match this result item. The detailed view should be replaced by the Map view, as this item will no longer be in the result list:
Screenshot 2024-09-20 at 5 12 42 PM

6. Search for "a" in the search box. You should see the "no results" view with all matched subject buttons displayed. Note that I’ve simply placed all subject buttons here for now; we can adjust this after improving the recommendation system:
Screenshot 2024-09-20 at 5 13 05 PM

7. Click "Compose," and you should see the same result as in step 2. During these operations, the user can always click "reset all filters" to return to the original state without interfering with the query results:
Screenshot 2024-09-20 at 5 15 39 PM


pengyin-shan:
The last commit removes the "Compose" icon tag, so please skip step 7 in the description above

mradamcox (approved):
Looks great, thank you!

PR #312, opened 2024-09-26
fix: refactored Post + Showcase to cut down on duplicated code

bodom0015:
## Problem
News (Post) + Showcase contain a lot of copy-pasted / duplicated code between them

We would like to consolidate this into a set of shared components that can be used for both contexts

## Approach
* Refactor `PostItem`, `PostList`, and `PostLayout` into reusable components
    * Replace copy-pasted code within `ShowcaseItem`, `ShowcaseList`, and `ShowcaseLayout` to use these newly-refactored components

## How to Test
TBD

### DIFF
--- a/src/components/news/PostItem.tsx
+++ b/src/components/news/PostItem.tsx
@@ -32,11 +32,13 @@ export default function PostItem({
   `;
 
   return (
-    <Link href={slugPrefix + item.slug} legacyBehavior>
+    <>
       <style>{css}</style>
-      <a className={"no-underline item-link"}>
-        {getContent(item)}
-      </a>
-    </Link>
+      <Link href={slugPrefix + item.slug} legacyBehavior>
+        <a className={"no-underline item-link"}>
+          {getContent(item)}
+        </a>
+      </Link>
+    </>
   );
 }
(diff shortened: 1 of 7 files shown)


mradamcox (changes requested):
@bodom0015 This looks good, I see where you are going with combining the ShowcaseLayout directly into the PostLayout, but because we are about to add a 3rd category, I'm wary of putting all of this conditional logic (conditional at an individual property-level, that is) into a single component. I think it could get messy and make future readability more difficult. So, would be able to make these changes:
1. Separate these back to standalone layouts for Posts and Showcase. I'd especially like to keep the prop lists separate, and they should effectively mimic the fields as defined in `/public/config.yml`
    It's definitely possible that some smaller components _within_ these layouts will be shared though, and certainly figuring out a way to combine the CSS would be good. Perhaps the `Layout.tsx` component that is in `news` should be abstracted up one directory and both ShowcaseLayout and PostLayout could conditionally rendered within it?
2. In `/lib` look at the two files that hold management functions for these objects, `showcases.js` and `posts.ts`, and clean these up a bit: Each should only reference its specific content type, and, again, think of how a `guides.ts` will fit into this setup.

Thanks!

bodom0015:
Closing this one instead of backing out individual changes, will reopen a new PR from a new branch :+1:

Issue #311, opened 2024-09-25
Add the new information area to the search box


pengyin-shan:
<img width="535" alt="Screenshot 2024-09-25 at 9 12 22 AM" src="https://github.com/user-attachments/assets/7233629a-50ca-4ce5-8c2b-1a6515b41be6">

PR #314, opened 2024-10-01
#311: Filter object from the solr side and add info section

pengyin-shan:
This PR addresses #314. It also moves the judgment of unsuppressed documents to the query side.

## How to Test

1. Navigate to the search page and click the info icon in the search box. You should see the information panel appear:

Screenshot 2024-10-01 at 9 16 34 AM

2. Switch tabs, and different content should appear. I have added placeholders for now.

3. Click on the info icon or the close icon. This tab should be replaced by the original search box.


mradamcox (approved):
Looks good, thanks!

Issue #317, opened 2024-10-03
Add "Guides" category to CMS


mradamcox:
We already have two categories of content in the CMS, Posts and Showcases, now we need a third, Guides. These are research guides on a specific SDOH topic. This will begin with a new entry in the `collections` section of the config YAML: https://github.com/healthyregions/SDOHPlace/blob/main/public/admin/config.yml.

The new collection will need these properties at least:

```
-  name: "guide" # Used in routes, e.g., /admin/collections/blog
    label: "Guide" # Used in the UI
    folder: "content/guides" # The path to the folder where the MDX files are stored
    extension: "mdx"
    format: "frontmatter"
    create: true # Allow users to create new documents in this collection
    slug: "{{slug}}"
```

The `fields` section can mimic the "showcase" collection generally (there are good examples of all fields there and in the posts collection), and these fields should be sufficient:

- `title`
- `last_updated` (date field)
- `featured_image`
- `author` (this should just be a string field for now, no need for a relation widget or anything like that)
- `body`

mradamcox:
Now completed and we have a first example guide up and linked to the homepage as well.

Issue #303, opened 2024-09-20
CMS: Add date field to Showcase pages


mradamcox:
I don't think we need to display this field anywhere, we just want it for sorting how the showcases are listed in the main /showcase page.

PR #320, opened 2024-10-07
fix: refactored lib folder patterns surrounding posts + showcases

### LINKED ISSUE #302: CMS: Audit logic/types around News and Showcase system
The News posts are collected and presented on the frontend with a set of code that I found in an example repo. When I created the Showcase category, I just duplicated much of this setup, but in some cases the lines may still be blurry, or there may be unnecessary duplicative code between the systems. This ticket is for a general audit and cleanup of these two systems, with an eye toward the fact that we will be adding a third category of content, Guides, in the relatively near future.

bodom0015:
## Problem
There is quite a bit of copy-pasted code between the Posts + Showcases

We would like a better pattern for adding new categories to avoid copy-pasting as much in the future

Redo of #312 

Fixes #302 

## Approach
* fix: restructured PostLayout + Showcase into `src/components/Layout`
* fix: refactor PostLayout + ShowcaseLayout slightly to more easily include new categories in the future (e.g. Guides, see #317)
* fix: refactor lib folder to cut down on some of the copy-pasted bits - maintain separate implementations for these resources
    * NOTE: After cleaning up this code, the Showcases are no longer sorted by Date - this will be fixed by #303 

## How to Test
1. Navigate to the list of "Project News" (PostList / PostItem): http://localhost:3000/news
    * You should see no changes compared to production: https://sdohplace.org/news
2. Choose an article (Layout -> PostLayout): http://localhost:3000/news/finding-healthy-food-connecticut
    * You should see no changes compared to production: https://sdohplace.org/news/finding-healthy-food-connecticut
    * NOTE: Images appear differently here, but these are not style changes - production appears to use different screenshot paths than dev
    * NOTE: Copyright symbol (at the bottom) has moved to the left a few pixels - I could not track down why, but this is barely noticeable
3. Navigate to the list of "Fellow Showcases" (ShowcaseList / ShowcaseItem): http://localhost:3000/showcase
    * You should see no changes compared to production: https://sdohplace.org/showcase
    * NOTE: Images appear differently here, but these are not style changes - production appears to use different screenshot paths than dev
4. Choose a showcase (Layout -> ShowcaseLayout): http://localhost:3000/showcase/windham-willmantic-conn-food-access-map
    * You should see no changes compared to production: https://sdohplace.org/showcase/windham-willmantic-conn-food-access-map
    * NOTE: Images appear differently here, but these are not style changes - production appears to use different screenshot paths than dev
    * NOTE: Copyright symbol (at the bottom) has moved to the left a few pixels - I could not track down why, but this is barely noticeable


src/components/Layout.tsx
```diff
@@ -0,0 +1,199 @@
...
+  news_props?: PostLayoutProps;
+  showcase_props?: ShowcaseLayoutProps;
+  /*guide_props?: GuideLayoutProps;*/
+  page_header?: string;
+  children?: React.ReactNode;
+};
+export default function Layout({ type, news_props, showcase_props, page_header, children }: Props) {
+  return (
+    <>
+      <Header title={"News"} />
```
mradamcox, line 20: This should reflect the content on the page and not always be "News" for all Layouts.

src/components/Layout.tsx
```diff
@@ -0,0 +1,199 @@
...
+import NavBar from "@/components/NavBar";
+import TopLines from "@/components/TopLines";
+import Footer from "./homepage/footer";
+import React from "react";
+import PostLayout, {PostLayoutProps} from './news/PostLayout';
+import ShowcaseLayout, {ShowcaseLayoutProps} from './showcase/ShowcaseLayout';
+
+type Props = {
+  type?: 'news' | 'showcase' /* | 'guide' */;
+  news_props?: PostLayoutProps;
```
mradamcox, line 11: Looks good to me, thanks for reworking these!

mradamcox (changes requested):
Just one small change needed, to update the title in the header, see comment.

src/components/Layout.tsx
bodom0015: Oooh, good catch! thank you, I can quickly fix this one :pray:

bodom0015: I've added a simple function with a `switch` statement here to render the title properly based on the provided `type`.

If we would prefer, I can write a quick function to split the type into words and capitalize the first letter of each word. I held off on this, since it didn't seem necessary, but if we would prefer this approach we can do that instead :)

bodom0015:
I'm noticing that the `font-size` of the page title has changed somewhere... checking to see if I may have lost a bit of needed CSS along the way 🤔

mradamcox (changes requested):
The title should be the same as the current site, before the refactor. Please look at production site and follow that. One place where I see it's currently wrong on the production site is that /showcase is "Showcase | News | SDOH & Place Project", but that should just be "Showcase | SDOH & Place Project". Perhaps an overhaul of titles across the entire website would be good in the future, but for now we just need to at least not regress with this refactor.

bodom0015:
@mradamcox I think I've fixed all the styling discrepancies here :+1: sorry for the additional delay on this - I completely agree that we don't want any styling changes here as a part of this merge

Regarding the `title`, it looks like this is set incorrectly in `config.json` (contains "News"). Should I go ahead and correct that here? I don't think this is a regression if the bug currently exists on the production site, but maybe I'm missing some context?

mradamcox (commented):
merged to main branch, thanks!

PR #350, opened 2024-10-24
feat: register a custom editor component named Toggle Section

### LINKED ISSUE #342: Create new "Collapsible" markdown editor component - CMS
In the Decap CMS markdown editor we would like to add a new block component option to handle collapsible sections. At the core of this component would be the standard HTML elements like this:

```
<details>
  <summary>text of the title</summary>
  Text of the detailed content
</details>
```

And it would behave like this:

<details>
  <summary>text of the title</summary>
  Text of the detailed content
</details>

There is no need to program any of the toggle behavior, that will come automatically, and we should have styles set up already for this as well. We just need to ease the creation of these styles of elements in 

Here is the Decap documentation about how to create and register a custom component: https://decapcms.org/docs/custom-widgets/. Note that we will need to `registerEditorComponent`, most likely in the public/admin/index.html file.

The name of the new component should be "Toggle Section". Inputs for the component should be

1. title (required)
    Will be placed within the `<summary>` tag and shown at all times
2. content (required)
    Will be placed within the content (after the summary tag), and will only appear when the toggle is open.
4. group (optional)
   This would be a single string that gets mapped directly to the `name` attribute of the `details` element. This provides a radio-type behavior for all details in the same group, if one is opened all the others will close (this is default HTML behavior, we only need to supply the name attribute. See example: https://nikitahl.com/native-html-accordion

For the content, _ideally_ we could be putting Markdown in here, and this would be rendered properly in the preview. If that proves problematic we could just leave plain text for now and come back to complex content later.

bodom0015:
## Problem
DecapCMS allows us to create custom editor components to render complex Markdown

Fixes #342 

## Approach
* feat: create a custom editor component that will produce standard `<details>`/`<summary>` toggles as HTML
    * Prevent nesting one Toggle Section within another Toggle Section

Example:
Screenshot 2024-10-24 at 1 26 29 PM

Which produces the following HTML:
```html
<details name="example"><summary>1. Access varies depending on where you are (origin)</summary><p>Imagine living in downtown Chicago, where it is easy to catch a bus or hop on the subway to get to work, visit friends, or run errands. If the transit system is well-connected, and stops are just a short walk from your home, this could make it convenient for you to rely on public transportation. Now, if someone lives in a small rural town where public transit might be limited to a few bus routes that only run a couple of times a day. The nearest bus stop could be several miles away, making it inconvenient to rely on public transportation. At the same time, some individuals may find it convenient to obtain services near their work or shopping locations, indicating that daily activity spaces other than residential locations can be more representative of origin in measuring public transit calculations. This highlights the importance of origin (location) in accessibility decision making.</p></details>
```

### TODOs
* ~~Newline characters within `content` are not well-supported. This will be problematic for less technical users (supporting which is the whole point of using a CMS system to begin with)~~ regex has been adjusted to allow newline characters within collapsible content :+1:
* ~~Images embedded in `content` do not play well with the editor - this may be because of the newline support mentioned above~~ with the corrected newline behavior (see above), this problem appears to have gone away and we can now embed images within collapsible content :+1:

## How to Test
1. Navigate to https://deploy-preview-350--cheerful-treacle-913a24.netlify.app/admin/index.html
2. Click on Guides on the left, and choose the "Public Transit Equity" guide from the list
    * You should see an Editor on the left, and a preview of the contents on the right
3. Ensure that the editor is set to "Rich-text" mode
4. Expand the "+" dropdown on the editor
    * You should see "Toggle Section" listed here - this is our newly-added component
5. Choose "Toggle Section" from the list
    * You should see a new widget named "Toggle Section" appear in the editor below
    * You should see that this widget has 3 separate inputs: `title`, `content`, `group`
6. Enter a value for `Title` - this will be the clickable header when the section is collapsed
7. Enter a value for `Content` - this will be Markdown that becomes the main content for when the section is expanded
    * Expand the "+" button here - you should see that Toggle Section is not listed as an option within a Toggle Section
8. (optional) Enter a value for `Group` - if specified, only a single section from the same group can be expanded at once
9. Change between the "Rich-text" and "Markdown" modes in the editor
    * You should see the widget is converted to its raw HTML format (see above) and then back again
10. Try to come up with weird edge cases that might break this widget
    * NOTE: The HTML form must be a produced as a single line, or else we have trouble detecting the end of one Toggle Section and the start of another

### DIFF
--- a/public/admin/index.html
+++ b/public/admin/index.html
@@ -29,6 +29,91 @@
         "https://fonts.googleapis.com/css2?family=Nunito:wght@300;400&display=swap"
       );
       CMS.registerPreviewStyle("/styles/posts.module.css");
+      // Register custom Editor Components
+      CMS.registerEditorComponent({
+        // Internal id of the component
+        id: "toggle-section",
+        // Visible label
+        label: "Toggle Section",
+        // Fields the user need to fill out when adding an instance of the component
+        fields: [
+          {
+            name: 'title',
+            label: 'Title',
+            widget: 'string'
+          },
+          {
+            name: 'content',
+            label: 'Content',
+            widget: 'markdown'
+          },
+          {
+            name: 'group',
+            label: 'Group',
+            widget: 'string'
+          }
+        ],
+
+        // Regex pattern used to search for instances of this block in the markdown document.
+        // Patterns are run in a multiline environment (against the entire markdown document),
+        // and so generally should make use of the multiline flag (`m`). If you need to capture
+        // newlines in your capturing groups, you can either use something like
+        // `([\S\s]*)`, or you can additionally enable the "dot all" flag (`s`),
+        // which will cause `(.*)` to match newlines as well.
+        //
+        // Additionally, it's recommended that you use non-greedy capturing groups (e.g.
+        // `(.*?)` vs `(.*)`), especially if matching against newline characters.
+        pattern: /^<details( name="(.*)?")?><summary>(.*?)<\/summary>(.*?)<\/details>$/m,
+        // Given a RegExp Match object
+        // (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#return_value),
+        // return an object with one property for each field defined in `fields`.
+        //
+        // This is used to populate the custom widget in the markdown editor in the CMS.
+        fromBlock: function(match) {
+          console.log('Matches: ', match);
+          return {
+            title: match[3],
+            content: match[4],
+            group: match[2]
+          };
+        },
+
+        // Given an object with one property for each field defined in `fields`,
+        // return the string you wish to be inserted into your markdown.
+        //
+        // This is used to serialize the data from the custom widget to the
+        // markdown document
+        toBlock: function(data) {
+          if (data.group) {
+            return `<details name="${data.group}"><summary>${data.title}</summary>${data.content}</details>`;
+          } else {
+            return `<details><summary>${data.title}</summary>${data.content}</details>`;
+          }
+        },
+        // Preview output for this component. Can either be a string or a React component
+        // (component gives better render performance)
+        toPreview: function(data) {
+          if (data.group) {return `
+<details name="${data.group}">
+  <summary>${data.title}</summary>
+
+  ${data.content}
+
+</details>
+`;
+          } else {
+            return `
+<details>
+  <summary>${data.title}</summary>
+
+  ${data.content}
+
+</details>
+`;
+          }
+
+        }
+      });
       // this preview generally mimics the structure of the actual post page, but it is not
       // actually derived from that component (tried for a long time to get that to work, no luck.)
       var PostPreview = createClass({
--- a/public/admin/index.html
+++ b/public/admin/index.html
@@ -63,7 +63,7 @@
         //
         // Additionally, it's recommended that you use non-greedy capturing groups (e.g.
         // `(.*?)` vs `(.*)`), especially if matching against newline characters.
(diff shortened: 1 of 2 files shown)


mradamcox:
Looking good so far! However, it seems like even though the Markdown widget is used for the content, it doesn't get rendered properly:

image

Better behavior would either

1. (preferred) render markdown to HTML in the preview as well as in the final page (I didn't check the final page, but assume it isn't working there either), or
2. (alternative) replace the markdown widget for the content field with a plain text widget.

I'm assuming we can write raw HTML into this content regardless of what widget is used, so if that is the only way we can have complex content, like lists or images, then we should force the user to do that, and not provide a markdown widget that doesn't work.

mradamcox (changes requested):
(see above comment)

bodom0015:
Thank you for the feedback, @mradamcox!

Sorry, I had based this implementation off of the DecapCMS example "Collapsible Note":
https://decapcms.org/docs/custom-widgets/#registereditorcomponent

They specified the widget type as `markdown`, so I assumed this would handle Markdown appropriately by default 🤔 
Apparently this case needed some special care and we now use `marked` to parse the Markdown within the Toggle Section:
https://github.com/markedjs/marked

Gotchas involved that we're working around here:
* there must be a space before markdown content or else it will not render properly - I think I've added this in the background so the user shouldn't need to worry about it
* single-quoted attributes appeared to work best with the RemoteMDX version on `/guides/[slug]`, even though both single and double quotes are accepted by the editor in DecapCMS
* regex needed some special care to properly detect the end of the value of the `name` attribute - without this there were some errors parsing group name out of adjacent Toggle Sections
* This does NOT support nested Toggle Sections, even though the widget is offered in the sub-editor as well - after converting to Markdown and back, the sub-sub-editors will be lost, as there is it is not trivial to parse out these nested values using regex

Things that I should be well tested:
* Adding a Toggle Section via the button in the Editor
* Converting entire document back and forth in Rich-text vs Markdown this transformation works both ways
* Inserting Markdown values into the `content` of the Toggle Section
* Creating / Converting Toggle Sections both with and without `group` specified
* All of the above verified on the resulting `/guides/` view as well (rendered via RemoteMDX)

mradamcox (changes requested):
Looking really good. Can you see about disabling the toggle section component in the nested markdown widget? https://github.com/decaporg/decap-cms/issues/3459. It would be good to know if it is possible in general, because we will likely end up with other custom components that we only want to have available in certain uses of the markdown widget.

bodom0015:
@mradamcox yes, it looks like we were able to prevent the user from adding one Toggle Section with an existing Toggle Section :+1:

```json
          {
            name: 'content',
            label: 'Content',
            widget: 'markdown',
            // exclude toggle-section from being nested inside of another toggle-section
            editor_components: [ "image", "code-block" ],
            required: true
          },
```

Adding the `editor_components` property allows us to customize which buttons are offered under the `+` button in the sub-editor :tada:

mradamcox:
Ok, perfect, thanks for figuring that out. It will make it a lot easier to add custom editor components and only use them in certain instances of the markdown widget. One _final_ request for this PR: Could you add `cursor: pointer` style to the summary element, so that the toggle looks a little bit more like a button when you hover over it? After that, ready to merge!

bodom0015:
@mradamcox awesome! I was able to add this style pretty easily into our rendered version of the CMS (e.g. `/guides/public-transit-equity`)

I didn't see an easy way to adjust this in the DecapCMS preview as well, but I can spend some time to look into that if it is important?

mradamcox (approved):
Looks good to me. As we get more familiar with this pattern we can probably drop a lot of these comments, but for now I think it's fine to retain them.

PR #352, opened 2024-10-25
Similar results short-term solutions

pengyin-shan:
This PR represents the front-end implementation of the 'Similar Result' section (#313). I will create a separate PR for the metadata manager to update the Solr configuration.

After multiple attempts, I found that the pre-filter for the 'suppressed' attribute in the `suggest` endpoint still not working. To optimize it as much as possible for now, I've implemented front-end adjustment: Once all of the suggestions (max 200) have been returned, the autocomplete will filter out terms with titles that include the '(year)' format, such as 'XXXIndex (2012),' since these are likely child records. 

This solution is not perfect, as some records may still appear, but we can improve it in the future.

Another challenge is the multiple asynchronous calls triggered by autocomplete suggestion calls, term search calls, and other searches that match each suggestion of the final term. These caused overwritten results and delays in presenting them. I updated the handleSearch mechanism to optimize this.

### How to Test

1. Navigate to the search page and type 'a'.  ~~You should now see all location-like terms have been moved to the bottom of the suggestion dropdown~~ (In my latest Solr configuration, I managed to remove location terms related to spatial coverage (such as "Illinois") from the Solr side, **so you should not see any spatial coverage data appear in the auto-complete now**) :
Screenshot 2024-10-28 at 11 47 03 AM

2. Press Enter to simulate the audience searching for 'a', which is a case expected to generate a large amount of related results. You should see a progress bar appear for a few seconds, followed by the recommended results appearing:

Screenshot 2024-10-28 at 11 47 54 AM

Note: I changed "Similar Results" to "You may want to check," as we're now adopting a more recommendation-style approach.

3. Choose the "average download speed" option from the dropdown. You should receive only one result since there is no additional suggestion for this query:

Screenshot 2024-10-28 at 11 50 30 AM

4. Search for "soc" without selecting anything from the dropdown. You should receive the "You may want to check" results faster since there are fewer suggestions to search for. **Using a higher weight could speed this up more (currently I use 50/100). Let me know if you want to use a higher bar**:

Screenshot 2024-10-28 at 11 52 45 AM

5. You should be able to review the details of a suggested item by clicking the "View" button on any object.

Screenshot 2024-10-28 at 12 14 25 PM


mradamcox (approved):
Nice work, this is coming along well. A few ideas:

1. Couldn't the `suppressed` field be used directly as a filter in the recommended results? Seems like it would be easy enough to ensure that it is returned with the result items and then used for this "post-filtering" instead of inferring from the titles? I may be missing something here though.
2. Having so many calls to the `select` endpoint seems like a pattern that may not be ideal down the road, so I'm wondering if either of these changes would reduce these calls:
    - Not providing suggested results at all if a query returns at least 10 results
    - Only providing a maximum of 4 suggested result items.
3. For now, let's change the text to "You may be interested in..." though I expect we'll have further updates to this phrase in the future.

pengyin-shan:
> Nice work, this is coming along well. A few ideas:
> 
> 1. Couldn't the `suppressed` field be used directly as a filter in the recommended results? Seems like it would be easy enough to ensure that it is returned with the result items and then used for this "post-filtering" instead of inferring from the titles? I may be missing something here though.
> 2. Having so many calls to the `select` endpoint seems like a pattern that may not be ideal down the road, so I'm wondering if either of these changes would reduce these calls:
>    
>    * Not providing suggested results at all if a query returns at least 10 results
>    * Only providing a maximum of 4 suggested result items.


> 3. For now, let's change the text to "You may be interested in..." though I expect we'll have further updates to this phrase in the future.

Hi @mradamcox, I’ve made the following changes:

1. Suggestions are now limited to 10 items, with recommended results also capped at a maximum of 10, ranked by highest weight.

2. The Solr configuration has been adjusted to include the suppressed field in the payload (unfortunately, I cannot concatenate title and suppressed together; this must be handled at the document level). As a result, most suppressed items should now be filtered out in suggestions, with only a few exceptions.

3. For extra-long search terms (e.g., when a user selects a long item from the dropdown that exceeds 50 characters), only the first five terms will be used in the query. This prevents Solr from struggling with overly lengthy queries while still producing similar results. This can be adjusted as needed:
Screenshot 2024-11-04 at 2 43 33 PM
Screenshot 2024-11-04 at 2 45 22 PM

mradamcox:
Ah, I understand now how you are needing to access the suppressed field during the suggestion, and I think the payload makes sense for now. Good to become more familiar with that response object anyway, maybe we'll be create a custom fields for a more complex payload in the future.

Overall this work looks good, I'll merge it now.

PR #359, opened 2024-11-05
Add theme filters to the filter panel

pengyin-shan:
This PR adds "Theme" filters to the filter panel, allowing users to directly explore different themes rather than only seeing them on the "No Results" page.

## How to Test
1. Go to the search page and open the filter panel. You should see a list of Theme filters:
Screenshot 2024-11-05 at 3 07 33 PM

2. Click on any theme icon. The results list should update to show only items related to that theme. Note that all other filter and search settings will be reset:
Screenshot 2024-11-05 at 3 07 59 PM


mradamcox (approved):
Looks good!

PR #361, opened 2024-11-07
feat: added "date" field to showcase items

### LINKED ISSUE #303: CMS: Add date field to Showcase pages

bodom0015:
## Problem
Showcase items cannot be easily sorted by date, since we are not tracking the dates they were added

Fixes #303 

## Approach
* feat: added a new `date` field to Showcase items
    * This adds a new `DateTime` widget to the Showcase editor (defaults to the current date, no time)
    * This `date` field is not currently displayed on the list of Showcases or when viewing a rendered Showcase :+1:
    * Sort ShowcaseList by this new date field (newest to oldest)
* NOTE: This PR does not retroactively add `date` fields to older Showcase items (but I am happy to add these if we know the expected/desired date for each Showcase)
    * ~~Old entries without a date are currently tolerated - no extra errors should occur if they are not present~~ date is now required, since we are sorting by it
    * To add a `date` to an old entry, you can simply use DecapCMS to Save/Publish a new version

## How to Test
1. Navigate to /admin/index.html
2. Click on Showcases on the left side
3. Choose a Showcase from the list to edit
    * You should be brought to the Showcase Editor
    * You should see a new widget about halfway down the page that says `date`
4. If you add a date and click "Save", a new PR is created that will add the date field to the Showcase item
    * For example: https://github.com/bodom0015/SDOHPlace/pull/2/files

### DIFF
--- a/content/showcase/sharing-birth-stories-in-the-motherhood-project.mdx
+++ b/content/showcase/sharing-birth-stories-in-the-motherhood-project.mdx
@@ -2,6 +2,7 @@
 title: "Sharing Birth Stories in The Motherhood Project "
 slug: sharing-birth-stories-in-the-motherhood-project
 image: /images/showcase-thumbnail-3-.png
+date: 8/14/2024
 link: https://storymaps.com/stories/e57626b6483e4ddfb35f2c06cd914bfd
 tech_used: Storymaps.com
 fellow: Makiya Thomas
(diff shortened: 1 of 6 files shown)


mradamcox (changes requested):
Looks good, but (and I realize the ticket didn't actually say this explicitly) could you add sorting on this new field to the Showcase list page? With the most recently dated showcase entries on top.

Also, it would be good to add dates to the four showcases that are already in this main branch. Here are dates you can use, and you can just put them directly in the frontmatter of the MDX files (no need to use Decap for this):
- mapping montgomery -- 7/23
- willamantic -- 7/26
- structural inequity -- 8/2
- motherhood project -- 8/14

Don't worry about the ones in the publish branch.

PR #366, opened 2024-11-11
feat: added Newsletter button to the footer

### LINKED ISSUE #47: Integrate Newsletter
Once a newsletter is created, we'll need to implement the subscription box on the homepage.

bodom0015:
## Problem
We don't currently have a link allowing the user to subscribe to the Newsletter / Mailing list

Fixes #47 

## Approach
* fix: use actual SVG in ButtonWithIcon (instead of ``)
    * This now uses Material UI's [button with label & icons](https://mui.com/material-ui/react-button/#buttons-with-icons-and-label) under the hood
    * This only affects ~6 buttons in the UI
* fix: customized SVG loader in webpack
    * This allows us to import a custom SVG and use it immediately as a React component
    * This only affects import URLs that end with `?component` (to preserve existing imports where we need the raw SVG - there are a lot of these)
* feat: added a Newsletter button to the header that opens the subscribe link in a new tab

Screenshot 2024-11-11 at 11 52 55 AM

## How to Test
1. Navigate to the homepage
2. Scroll down to the footer
    * You should see a NEWSLETTER button now appears here
3. Click the NEWSLETTER button
    * You should see a page open in a new tab that allows you to subscribe to the newsletter / mailing list

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -17,3 +17,8 @@ declare module "*.yml" {
   const value: any;
   export default value;
 }
+
+declare module "*.svg?component" {
+  const value: any;
+  export default value;
+}
--- a/next.config.js
+++ b/next.config.js
@@ -8,16 +8,30 @@ module.exports = {
     unoptimized: true,
   },
   webpack: (config, { buildId, dev, isServer, defaultLoaders, webpack }) => {
+    // Grab the existing rule that handles SVG imports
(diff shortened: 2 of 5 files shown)


mradamcox (approved):
Splendid, thank you! And thanks for fixing up the icon pattern, we'll be sure to migrate toward this moving forward.

Issue #363, opened 2024-11-07
Enable theme search in the filter panel to support multiple searches


pengyin-shan:
OR relation

PR #370, opened 2024-11-11
enable mul-ti selection of the theme filter

pengyin-shan:
This PR addresses #363. It enbles multi-selection for the themes filter.

##How to Test
1. Navigate to the search page and open the filter panel.

2. Select the "Demographic" theme filter, then add the "Employment" filter. You should see the result list length increase due to the additional filter:
Screenshot 2024-11-11 at 12 46 17 PM

3. Open a result detail, and the corresponding theme icon should be highlighted:
Screenshot 2024-11-11 at 12 45 31 PM

4. Experiment with the filters to ensure the list updates as expected.

### DIFF
--- a/src/components/search/detailPanel/detailPanel.tsx
+++ b/src/components/search/detailPanel/detailPanel.tsx
@@ -79,7 +79,6 @@ const DetailPanel = (props: Props): JSX.Element => {
                 labelColor={fullConfig.theme.colors["almostblack"]}
                 roundBackground={true}
                 handleSearch={props.handleSearch}
-                handleInputReset={props.handleInputReset}
               />
             ))}
           </div>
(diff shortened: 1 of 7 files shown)


mradamcox (changes requested):
Great, works as expected. Only thing we need is change the enabled color on border, icon, and text to french violet.

pengyin-shan:
> Great, works as expected. Only thing we need is change the enabled color on border, icon, and text to french violet.

Did you mean to change the color to French violet after the icon is selected?

mradamcox:
| Did you mean to change the color to French violet after the icon is selected?

Yes, when a filter is enabled change the colors to French violet. For the default colors, leave as is.

pengyin-shan:
> | Did you mean to change the color to French violet after the icon is selected?
> 
> Yes, when a filter is enabled change the colors to French violet. For the default colors, leave as is.


I updated the code to change the color when a theme is selected:
Screenshot 2024-11-12 at 1 26 17 PM

Issue #376, opened 2024-11-13
Add better containers within Discovery App layout


mradamcox:
I believe it will be beneficial to wrap the main elements of the discovery app in container elements that will constrict the width of the content on wider screens. Zoomed out, for example, it currently looks like this
![image](https://github.com/user-attachments/assets/7738e088-8101-4479-a66f-055a004f7e23)
but having better control over the inner elements should easily make that look better.

PR #378, opened 2024-11-13
style: add containers around central content in deiscovery app #376

mradamcox:
This PR adds two internal containers around the content within the discovery app search bar, and the middle content below. This isn't a significant change, but will help the layout stay contained on very wide screens, and I believe create a little better foundation for improving the mobile view down the road.

### DIFF
--- a/src/components/search/detailPanel/headerRow/headerRow.tsx
+++ b/src/components/search/detailPanel/headerRow/headerRow.tsx
@@ -75,7 +75,7 @@ const HeaderRow = (props: Props): JSX.Element => {
           </div>
           <Box
             sx={{ display: "inline" }}
-            className={`text-4xl leading-8 ml-[0.5em] ${classes.introCard}`}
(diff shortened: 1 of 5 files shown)


pengyin-shan (commented):
I tested by zooming out to simulate a larger screen, and everything appears to display well.

Just have one question. On large and medium-sized screens, it seems that the left padding on the page has disappeared and the padding in the middle is less than before. Here is what I'm seeing on my monitor:

Current dev site (1920 X 1080):
Screenshot 2024-11-13 at 4 59 52 PM

This PR (1920 X1080):
Screenshot 2024-11-13 at 5 01 03 PM

mradamcox:
Good catch on the left padding, I've adding another container in the middle section to properly match the banner section. You're right that the center gap is smaller, but looking back at the original Figma designs, it's actually even smaller there:
image
So I think this small change should be fine.

PR #377, opened 2024-11-13
fix: Address all feedback from Design Audit 11/11/2024

bodom0015:
## Problem
We've noticed some styling issues with the current SDOH website

This feedback has been recorded in the [Design Audit 11/11/2024 notes in Notion](https://www.notion.so/UX-Audit-2024-25-2a26f4fcfc394fe28150b5f26caacddc)

## Approach
* fix: Navbar (Desktop) - added SDOH Logo to desktop navbar in all views except for Home (logo is already displayed there)
* fix: Navbar (Mobile) - left-align menu options here, add Logo to all views
* feat: Navbar (Mobile) - added animation to the Menu/Close icon when it transitions from one to the other
* fix: Navbar dotted lines - fixed aspect ratio for `line1`, adjusted width of `line4` to match the others
* fix: Navbar - added additional space between each nav button, move navbar down slightly, and moved all views down slightly to accommodate this change
* fix: Navbar - added some extra padding between dropdown button and the menus that open when they are clicked
* feat: Navbar - added `box-shadow` to dropdown menus
* fix: Homepage - add some padding beneath the left side large logo to better align it with the right side
* fix: Homepage (Mobile) - adjust margins/padding to better balance the whitespace in this view, decrease total height by `1%` so that first panel fits on the screen (iPhone 12)

## How to Test
1. Navigate to homepage: https://deploy-preview-377--cheerful-treacle-913a24.netlify.app/
    * You should see the updated styles here for the Desktop Navbar and Desktop Homepage
2. Switch to mobile view
    * You should see the updated styles here for the Mobile Navbar and Mobile Homepage
3. Check other views to make sure nothing that drastic has changed
    * You should see that all views have had their content moved down slightly to accommodate the Navbar changes
    * You should see that SDOH Logo now appears in the Desktop Navbar on all views except for the Homepage

### Screenshots (Desktop)
Screenshot 2024-11-13 at 12 51 56 PM

Screenshot 2024-11-13 at 12 52 11 PM

### Screenshots (Mobile)
Screenshot 2024-11-13 at 12 51 39 PM

Screenshot 2024-11-13 at 12 51 45 PM

PR #380, opened 2024-11-14
UX Design Audit 11/14/2024

bodom0015:
## Problem
We would like to make some last-minute UX improvements before user testing begins

## Approach
NOTE: This branch contains all changes from #377 and #378 as well

Met with Shubham a few times this week for design audit / live-coding sessions

Notes can be found here with full list of changes: https://www.notion.so/UX-Audit-2024-25-2a26f4fcfc394fe28150b5f26caacddc

## How to Test
1. Navigate to https://deploy-preview-380--cheerful-treacle-913a24.netlify.app/search
    * You should see the SDOH&Place Discovery App
    * You should see that the list of improvements above have been implemented


mradamcox (approved):
These updates all look great to me. One thing I noticed is that the year filter only goes down to 2005 where it used to go to 1963. I've figured out this isn't an issue with the frontend, there is a mix of `int` and `str` values in the metadata records within the `index_year` field, causing the sorting of all values to return the wrong "oldest" date. I will fix this on the metadata side, so no problem here.

Issue #381, opened 2024-11-14
[Discovery App] Search bar loses focus after existing text removed


mradamcox:
This is a minor but definitely should be fixed ASAP.

1. Type text into the search bar
2. Hit backspace to remove all text
3. Try to type something else, and you'll see the input has lost focus: you need to click back on the bar in order to begin typing again.

Browser: Firefox & Chrome

PR #383, opened 2024-11-14
#381: Force the search box to refocus after the input got cleared out

pengyin-shan:
This PR addresses issue #381. It forces the search text box to refocus after the user removes all their input. I also removed some unused legacy code from discoveryApp.tsx.

##How to Test
1. Navigate to the search page and start typing in the search box.
2. Delete all input using the keyboard, and you should see that the search box remains focused after everything is cleared.


mradamcox (approved):
Looks good and works

Issue #382, opened 2024-11-14
Variables and labels within Details component


mradamcox:
There are a few small changes @Makosak and I had discussed that we should implement on the frontend.

1. We have added new field called "Featured variable" which is applicable to a handful of different metadata records.
    - In Solr, this variable is named `sdoh_featured_variable_s` and it's alias is `featured_variable` (I don't recall if aliases are used on the frontend, but just in case).
    - Please add this variable with label "Featured variable:" above the Methods Variables and Data Variables in the item details component--only show it if a value is present in the record.
3. Change label of "Methods Variable" to "Measures and data input:"
4. Change label of "Data Variables" to "What's in this dataset..."

Once we have a PR for this, we can solicit feedback for these labels from others, and perhaps they will be updated a bit.

pengyin-shan:
@mradamcox , I noticed that the sdoh_featured_variable_s field is missing in both aardvark_schema.json and sdohplace_schema.json. I can add this field to sdohplace_schema.json. Just to confirm, should it be similar to:
```javascript
  data_usage_notes": {
    "label": "Data Usage Notes",
    "uri": "sdoh_data_usage_notes_s",
    "data_type": "string",
    "multiple": false,
    "obligation": "optional",
    "controlled": false,
    "controlled_options": [],
    "schema": "SDOH Place",
    "display_group": "SDOH Place Project",
    "slug": "data-usage-notes",
    "widget": "text-area.html"
  }
``` 
right?

mradamcox:
Ah, good catch. I forgot we have those schema files in this repo as well. Are they used for lookups? Here is the current schema file that we use in the metadata manager: https://github.com/healthyregions/SDOHPlace-MetadataManager/blob/main/manager/metadata/schemas/sdohplace.json.

Honestly it would be best to perhaps download that file into this repo at build time, so are not storing two versions of the schema in different places. Also, I'm sure we are no longer (or shouldn't be, at least) using the aardvark_schema.json file in this repo, and it can be removed.

pengyin-shan:
Ah I see. Yes, there's a lookup to retrieve the corresponding variables from the schema file in the metadata folder. I can change it to fetch from https://github.com/healthyregions/SDOHPlace-MetadataManager/blob/main/manager/metadata/schemas/sdohplace.json if necessary, although it may introduce some data loading time. I don't think any modifications to the Solr object are needed, as all previous attributes are still present.

mradamcox:
[...]

Issue #375, opened 2024-11-13
Replace the 'Subject' part on the detail panel with another attribute


pengyin-shan:
Since we have confirmed that the Theme icons are indeed the "subject" of the result items, we need to replace the "subject" field with a different data attribute and potentially adjust the display order:
![Screenshot 2024-11-13 at 10 28 07 AM](https://github.com/user-attachments/assets/761c5d2f-b4b0-47e6-97ba-889994417b8d)

@mradamcox  @shubhamk008, what are your thoughts?

pengyin-shan:
Will change the 'Subject' to 'Keyword' as discussed in yesterday's meeting

PR #384, opened 2024-11-15
#382, #375: update details panel to add corresponding fields

pengyin-shan:
This PR addresses issues #382 and #375. It aims to add featured variables to the detail panel, switch "subject" to "keywords," and update method names.

## How to Test
1. Navigate to the search page and review the list. The term "subject" should be replaced with "keywords" in both the result card and the detail panel:
Screenshot 2024-11-15 at 12 03 20 PM

2. Search for "CDC Social Vulnerability Index."

3. Verify that the "featured variables" field is displayed, and the Methods/Data variable names have been updated:
Screenshot 2024-11-15 at 12 01 10 PM


pengyin-shan:
For this PR, I simply copied and pasted the latest SDOH JSON file. In the future, we could either fetch it from GitHub (or another location) or use a separate Git repository to automatically pull it during deployment.

mradamcox (approved):
Great, again it's possible that these labels will change further, but this is a big improvement for the time being.

PR #399, opened 2024-12-02
[WIP] Redux handling (in process)

pengyin-shan:
Draft PR, not buildable.


mradamcox (changes requested):
Overall, this is fantastic, and I'm super happy to have everything refactored like this. It will provide a much better foundation for work moving forward.

There are a few regressions in the page style and behavior, but I'm ready to merge this after only the following changes have been made. Other issues will be fixed with a more comprehensive planned update to the map component.
- Fix style of year filter
  - need to change the color and font, to go from this:
    image
  back to this:
    image
- Unchecking the "search this area" box doesn't refresh the search
- Fix vertical alignment of "search this area" box to match other buttons (we'll be adding another box in this row, so we should have a good pattern for identical arrangement here)

pengyin-shan:
> Overall, this is fantastic, and I'm super happy to have everything refactored like this. It will provide a much better foundation for work moving forward.
> 
> There are a few regressions in the page style and behavior, but I'm ready to merge this after only the following changes have been made. Other issues will be fixed with a more comprehensive planned update to the map component.
> 
> * Fix style of year filter
>   
>   * need to change the color and font, to go from this:
>     image
>     back to this:
>     image
> * Unchecking the "search this area" box doesn't refresh the search
> * Fix vertical alignment of "search this area" box to match other buttons (we'll be adding another box in this row, so we should have a good pattern for identical arrangement here)

Hi @mradamcox, to clarify our needs, should unchecking the "Search this area" box make the result list stay the same? In other words, if the user unchecks the "Search this area" box, does that mean they want to stop searching? If so, there should be no fetching actions triggered from that point onward until the box is checked again.

mradamcox:
Sorry about that, I was unclear. Currently, if you open the page there are 58 results, and clicking the "search this area" box reduces that number 47 (good). However, when I uncheck the box a new search isn't not triggered (which would bring the results back to 58), but it should be.

This checkbox is just another filter like Spatial Resolution or Theme, so check/uncheck should trigger new result retrieval.

mradamcox (approved):
Great, all looks good to me now. I may change that wording back on the map filter, but will wait until the dropdown is created so we can see how the labels and behavior all work together.

Thank you so much for this refactor! Looking forward to working with it next week.

PR #404, opened 2025-01-06
feat: added DecapCMS collection for Research Rabbit entries

### LINKED ISSUE #401: Create new CMS category for Research Rabbit collections
Lab members will soon begin creating public "collections" in [Research Rabbit](https://researchrabbitapp.com/), which will be focused on specific metadata records within the data discovery application. These collections can then be linked from within the item details section of the metadata app, which will make it easy for users to connect datasets in the system with literature that has used them.

To support this, we will make a new collection in our CMS. Here's how I'm thinking it will be configured (I don't remember every detail of how Decap will handle this structure but this should be enough):

- Collection name: "Research Rabbit"
- It will edit/create a single JSON file, similar to how the People collection is set up (though the people collection seems to have subcategories)
- The JSON file should be placed in `meta/`
- Each item in the collection will have the following fields:
  - Title (string)
  - Metadata record ID (string, multiple allowed)
  - Public link (URL)
  - Collected by: (string)*
  - Reviewed by: (string)*
  - Documentation/Report: (boolean)
  - Original peer-reviewed article by authors: (boolean)

*Decap _does_ allow us to make relationships between records, and we have done that a little bit already, so ultimately it would be really nice to make these fields link to people in either the Authors collection or Team collection, etc. But I think we should overhaul that all at once sometime in the future, because it would be good to fix this on a couple of other collection configs as well, so for now let's leave these fields as strings.

bodom0015:
## Problem
We would like to add a new collection for tracking Research Rabbit entries

Fixes #401 

## Approach
* feat: added a new collection in DecapCMS for Research Rabbit entries
* fix: move to Meta collection, allow multiple entries to be saved 

## How to Test
1. Navigate to `/admin/index.html#/collections/research-rabbit/entries/research-rabbit`
2. Click on the Meta collection
    * You should see a new listing here for Research Rabbit
3. Click on Research Rabbit
    * You should see a list of existing Research Rabbit entries listed here
    * You should see a button to "Add entry"
4. Click on "Add entry"
    * You should see a new Research Rabbit entry is added to the collection
5. Fill out necessary fields, click Save, then refresh the page
    * You should see that your changes are persisted
    * You should see that your changes are saved to the `meta/research-rabbit.json` file

### DIFF
--- a/public/admin/config.yml
+++ b/public/admin/config.yml
@@ -435,3 +435,46 @@ collections:
       - label: "Body"
         name: "body"
         widget: "markdown"
+
+  - name: "research-rabbit"
+    label: "Research Rabbit"
+    delete: false
+    editor:
+      preview: false
+    files:
+      - name: research-rabbit
+        label: Research Rabbit
+        file: "meta/research-rabbit.json"
+        label_singular: "item"
+        fields:
+          - name: title
+            label: Title
+            widget: string
+          - label: "Metadata Record ID"
+            name: "record_ids"
+            widget: "list"
+            summary: "{{record_id}}"
+            fields:
+              - label: Record ID
+                name: record_id
+                widget: string
+          - label: Public Link
+            name: link_url
+            widget: string
+            pattern:
+              [
+                '[(http(s)?):\/\/(www\\.)?a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)',
+                "Must be a valid URL.",
+              ]
+          - label: Collected by
+            name: collected_by
+            widget: string
+          - label: Reviewed by
+            name: reviewed_by
+            widget: string
+          - label: Documentation/Report
+            name: doc_report
+            widget: boolean
+          - label: Original peer-reviewed article by authors
+            name: peer_reviewed
+            widget: boolean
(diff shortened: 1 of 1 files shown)


mradamcox (changes requested):
@bodom0015 Looks good, just a couple of things to change.
1. In the config, can you move this outside of the Meta collection so it is its own collection? The file should still be `/meta/research-rabbit.json` but structurally I'd like to have Research Rabbit on its own.
2. I believe help text to each field so it appears under the data entry widget. For the "Record ID" field, could you add help text along the lines of "go to sdohplace.org/search, find and select a record, then copy the id from the URL (it will look something like `herop-jxbloa`)"

@marynia, can you add some pointers here for help text for the other fields in the list? Not every field needs one, but I think it would be nice to have for some less obvious ones like "Documentation/Report".
image

Makosak:
- Public Link: _Link to Research Rabbit URL for specific dataset of interest_
- Collected By: _Person who developed and curated the original Research Rabbit collection_
- Reviewed By:  _Person who reviewed the curated Research Rabbit collection_
- Documentation/Report: _Some documentation/report exists that describes how the dataset was developed. This is "gray literature" that was not peer-reviewed._
- Original Peer-Reviewed Article: _The authors of the dataset (or index featured by the dataset, for example) published a peer-reviewed article that details the development of the data of interest. This is linked in the Research Rabbit collection_

Issue #412, opened 2025-01-15
Go back to map view if user start a new search or clear out an existing search


pengyin-shan:
There have been cases where, if a user opens a detail view, then conducts a search, and the previous result is not included in the new list, the detail panel becomes empty. To resolve this issue, force the view to switch back to map view when a user conducts or clears a search

Issue #409, opened 2025-01-09
Create spellcheck feature to enable a FuzzySearch-style experience


pengyin-shan:
Enable Spellcheck so that when the user searches for a term that might have mis-spelling on the front-end, with Solr adjustment in https://github.com/healthyregions/SDOHPlace-MetadataManager/issues/35

Issue #405, opened 2025-01-07
Add highlight support to suggestions and search results


pengyin-shan:
After completing the Solr-side adjustments ([issue 33](https://github.com/healthyregions/SDOHPlace-MetadataManager/issues/33)), add front-end support for search term highlighting in both suggestions and search results. This will help users better understand the search result mechanism without opening the result card.

PR #413, opened 2025-01-15
#405, #409, #412: add support for highlight, spellcheck and better score mechanism

pengyin-shan:
This PR addresses #405, #409, and #412, reflecting front-end updates that align with changes made in https://github.com/healthyregions/SDOHPlace-MetadataManager/pull/37. Its main features include:

1. Enhanced Scoring Mechanism
A more scientific approach has been implemented to select items for the result list. Instead of relying on a fixed score threshold, a method was added to dynamically detect sharp score drops among results. Only items with acceptable score drops are included, while the rest are filtered out.

2. Tooltips for Transparency
Tooltips now display highlights from key fields, along with the score and a brief description of relevance. For related results, tooltips also indicate the targeted term for the result, helping users understand when a different but relevant term is suggested. This feature aims to enhance transparency in the search process.

3. Basic Spellcheck Feature
A basic spellcheck feature has been added to autocorrect misspelled terms and conduct searches with the corrected terms. Note that this is an initial implementation; some terms may still not be autocorrected, and further improvements are planned.

4. Some Bug Fixes

## How to Test

1. Navigate to the search page and browse through the items without performing a search; no tooltips should appear.

2. Test Autocorrect
    - Enter 'sdox' in the search box and either press Enter or click the search button.
    - The term should autocorrect to 'sdoh', and a small notification box will indicate that autocorrection has occurred.
Screenshot 2025-01-15 at 3 11 03 PM

3. Search for the Original Term
    - Click 'Search instead for sdox' to override the autocorrelation.
    - The "No results" view should appear, allowing users to search for the original term if needed.

4. Hover Over Search Results
    - Re-enter 'sdox' to trigger a search for 'sdoh'.
    - Hover over each search result to see:
        a. The score and search terms, showing how relevant the result is to the searched term.
        b. Highlights of the searched or related terms in artificial fields (e.g., 'SDOH' is related to 'Social Vulnerability'). Items with related terms may be ranked lower than those with exact matches in fields like Title or Description.
Screenshot 2025-01-15 at 3 39 47 PM
Screenshot 2025-01-15 at 3 39 59 PM

5. Test the Map and Detail View
    - Select a result to open the detail view, then perform another search.
    - The map view should replace the detail view to prevent a bug where the detail view becomes blank if the selected item is not in the new search results.
Screenshot 2025-01-15 at 3 39 14 PM

### DIFF
--- a/meta/helper/solrObjects.ts
+++ b/meta/helper/solrObjects.ts
@@ -32,6 +32,7 @@ const initSolrObject = (rawSolrObject: any, schema: {}): SolrObject => {
           return year.toString();
         })
     : [];
+  result.highlights = rawSolrObject.highlights ? rawSolrObject.highlights : [];
   result.meta = {};
   result.years = [];
   if (rawSolrObject.dct_isVersionOf_sm)
@@ -53,7 +54,7 @@ const initSolrObject = (rawSolrObject: any, schema: {}): SolrObject => {
       result.meta[schemaMatch(key, schema)] = rawSolrObject[key];
     }
   });
-
+  result.q = rawSolrObject.q;
   return result;
 };
 
--- a/meta/interface/SolrObject.ts
+++ b/meta/interface/SolrObject.ts
@@ -4,6 +4,7 @@
  * Note that not all attributes in metadata managers are here. Only ones returned by search results are included.
  */
 export interface SolrObject {
+  spellcheck: any;
   id: string;
   title: string;
   creator: string[];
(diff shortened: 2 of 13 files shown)


mradamcox (approved):
Looks great all around. @bodom0015 and I looked over all the pieces and makes sense to us. One thing we noticed is it looks like `CleanHtml.tsx` has `stripHtmlTags()` that that function is imported somewhere but never used. For that, I'll make a separate cleanup ticket that we can address later, not a big deal right now.

pengyin-shan:
> Looks great all around. @bodom0015 and I looked over all the pieces and makes sense to us. One thing we noticed is it looks like `CleanHtml.tsx` has `stripHtmlTags()` that that function is imported somewhere but never used. For that, I'll make a separate cleanup ticket that we can address later, not a big deal right now.

Ah good catch! This was used for suggestion highlighting, which I tried before but later removed because it was somewhat unnecessary and added extra processing time. I’ll clean it up later.

PR #415, opened 2025-01-16
408 display spatial coverage

### LINKED ISSUE #408: Add better mechanism for showing coverage of a given dataset within the map
Previously, we had a UI behavior where hovering the cursor over an item in the results panel would add a layer to the map showing the boundaries of that dataset. For example, if it was a dataset of census tract-level data, hovering the item would add a "highlight" layer of all census tracts to the map.

There were a few issues with this:
1. It wasn't an obvious behavior so users didn't necessarily understand what was happening.
2. Even though the highlight layer would load "quickly", it wasn't instantaneous, so if a user quickly hovered the mouse over an item and then moved away, the layer may not even load in that short amount of time.

I think it is worth trying a more explicit approach, so I propose that we add a checkbox within each item in the results panel labeled something like: "show spatial coverage". Checking this box would activate the highlight layer as before, but there would be no mystery for users as to how to visualize the spatial coverage of a dataset on the map.

As for how to implement this:

We already have pmtiles layers available for each spatial resolution (state, county, zcta, tract, block group) so we'll be able to utilize those layers as before. Further, we have the "highlight ids" fields which can hold multiple HEROP_ID values that can be used as filters against the pmtiles layers.

mradamcox:
This PR adds a new feature that allows users to check a box on each result item to add a preview layer to the map that shows the spatial coverage for a given dataset.

Closes #408 

We had something like this before on hover of the result item, but this is more explicit, and the implementation also offers a lot more flexibility in how the preview layer is constructed.

In the metadata schema we have a field "Highlight IDs" that must be filled out with one or more "HEROP_IDs", and that field powers this preview. The first three characters of a HEROP_ID denote what geography level the rest of the ID pertains to, and this is used to denote what geography layer is used for the preview layer. For a full explanation of HEROP_IDs, see: https://github.com/healthyregions. There are two ways to fill out the field.

1. With a * wildcard to show all geographies matching a pattern
2. With a list of specific HEROP_IDs to show a certain set of geographies

Examples of wildcards:

- All states: `040US*`
- All counties within Illinois: `050US17*`
- All ZCTAs: `860US*`
- All tracts in Cook County, Illinois: `140US17031*`

Examples of individual ID lists:

- Only show the state of Illinois: `040US17`
- Show counties around Chicago:
    ```
    040US17089
    040US17043
    040US17031
    040US17093
    040US17197
    ```

Here is a mockup example showing all counties in Illinois (though this dataset actually has a wider coverage area)
image

Big thanks to @bodom0015 for working through all of this with me!

### DIFF
--- a/meta/interface/SolrObject.ts
+++ b/meta/interface/SolrObject.ts
@@ -32,7 +32,7 @@ export interface SolrObject {
     suppress?: boolean;
     spatial_resolution?: string[];
     spatial_coverage?: string[];
-    sdoh_highlight_ids_sm?: string[];
+    highlight_ids?: string[];
     data_usage_notes?: string;
     data_variables?: string[];
     methods_variables?: string[];
--- a/src/components/map/dynamicMap.tsx
+++ b/src/components/map/dynamicMap.tsx
@@ -37,7 +37,7 @@ interface Props {
 
 export default function DynamicMap(props: Props): JSX.Element {
   const dispatch = useDispatch<AppDispatch>();
-  const { bbox, visLyrs, visOverlays } = useSelector(
+  const { bbox, visOverlays } = useSelector(
     (state: RootState) => state.search
   );
   const mapPreview = useSelector((state: RootState) => state.ui.mapPreview);
@@ -123,23 +123,7 @@ export default function DynamicMap(props: Props): JSX.Element {
         map.removeLayer(overlayRegistry[lyr].spec.id);
       }
     });
-    visLyrs.forEach((lyr) => {
-      if (
-        layerRegistry[lyr] &&
-        !mapLyrIds.includes(layerRegistry[lyr].spec.id)
-      ) {
-        map.addLayer(layerRegistry[lyr].spec, layerRegistry[lyr].addBefore);
-      }
-    });
-    Object.keys(layerRegistry).forEach((lyr) => {
-      if (
-        mapLyrIds.includes(layerRegistry[lyr].spec.id) &&
-        !visLyrs.includes(lyr)
-      ) {
-        map.removeLayer(layerRegistry[lyr].spec.id);
-      }
-    });
-  }, [visLyrs, visOverlays, mapLoaded]);
+  }, [visOverlays, mapLoaded]);
 
   const onMouseMove = useCallback((event: MapLayerMouseEvent) => {
     if (!mapRef.current) {
--- a/src/components/map/dynamicMap.tsx
+++ b/src/components/map/dynamicMap.tsx
@@ -59,7 +59,7 @@ export default function DynamicMap(props: Props): JSX.Element {
         "040": "state-2018",
         "050": "county-2018",
         "140": "tract-2018",
-        "150": "blockgroup-2018",
+        "150": "bg-2018",
         "860": "zcta-2018",
       }
       const source = lookup[previewLyr.filterIds[0].slice(0,3)]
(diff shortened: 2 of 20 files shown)


pengyin-shan (commented):
This is great! However, while testing https://deploy-preview-415--cheerful-treacle-913a24.netlify.app/search, I noticed that all the 'Show coverage area' options are greyed out (i.e., not clickable).

The Spatial Resolution filter's visualization on the map is also unavailable:
Screenshot 2025-01-17 at 10 21 41 AM

Is this the intended behavior? Or are there any data-related changes, such as re-indexing, that we need to implement?

I believe this PR is using `blacklight-core-dev` core.

mradamcox:
@pengyin-shan Sorry about that, I should have had a note, I will add `highlight_ids` values for a lot of the records right now so the boxes will be enabled soon.

Also: I decided to remove the spatial resolution `visLyrs` visualization because I'm not convinced it had really be helpful... We should discuss in with others and that can be added back in pretty easily if that's what we decide to do. We can compare this deploy preview against the branch build for the time being.

mradamcox:
@pengyin-shan One thing we noticed while working on this: the initial sort of records when you load the page seems to be not alphabetical (which I thought it was previously) but by last updated date (or something). In other words, if I update the top record, it will appear at the end of the list next time the page is loaded fresh. Do you think you could you set solr to return alphabetical by title on the initial load?

pengyin-shan:
> @pengyin-shan Sorry about that, I should have had a note, I will add `highlight_ids` values for a lot of the records right now so the boxes will be enabled soon.
> 
> Also: I decided to remove the spatial resolution `visLyrs` visualization because I'm not convinced it had really be helpful... We should discuss in with others and that can be added back in pretty easily if that's what we decide to do. We can compare this deploy preview against the branch build for the time being.

Ah I see. So, I assume both types of coverage will work later. I noticed the removal of `visLyrs`, and that’s fine on my end since we can use a single key for visual coverage for both Spatial Resolution and the result item's spatial coverage.

mradamcox:
The one thing I want to change about this before merging is to be able to include multiple wildcard ids--because currently only the first item in the list is evaluated if it is shown to have a wildcard in it, meaning there is no clean way to say all counties a few different states, without just making a list of each individual county id.

I'm thinking that the `any` expression we found could be used, and then we would use the slice matching for terms with `*` in them, or just direct matching for ids that don't have `*` in them--no need for the "in" array expression we were using before. @bodom0015 do you think you could try tackling that next?

The goal would be that a record with these three highlight ids

```
050US17*
050US55*
050US19067
```

would show all of the counties within Illinois (17), all counties in Wisconsin (55), and one single county in Iowa (19067).

mradamcox:
@bodom0015 Ok, one final thing I noticed while testing this feature out is that we still haven't managed a comprehensive reset of preview layer in all cases that need to be handled (there are more than I thought at first).

We need to remove the layer from the map anytime the search results are refreshed or cleared. Ways this can happen in the app are:

1. submit a query in the main search box
2. click "x" in the main search box to clear a query
3. enable or disable a filter
4. click the "clear all" button when a filter has been applied
5. select an entry in the location search box (located on the map)
6. clear a selected entry in the location search box

I think that's all, but maybe there is somewhere else. I remember explicitly handling a few of these examples, by setting the global preview list to `[]` but at least 5 and 6 aren't handled.

The solution I would like is to move this state change much closer to the single/main query function, perhaps baked right within it so every new query resets that list to `[]`. After that, I think we can merge this in...

mradamcox:
Added a few final update to this, a very light transparent fill to the preview layer, which better allows for the indication of a missing unit that is surrounded by included units:

image

Also fixed logic in the basic "negative" filter.

PR #417, opened 2025-01-17
#416: remove the initial state on sort

pengyin-shan:
This PR addresses issue #416 and removes the default sorting configuration from the front-end. The corresponding Solr configuration is available in [SDOHPlace-MetadataManager PR #39](https://github.com/healthyregions/SDOHPlace-MetadataManager/pull/39).

## How to Test
1. Navigate to search page. You should see that results are sorted by title by default.
Screenshot 2025-01-17 at 3 13 12 PM

2. Perform any search. Results should now be sorted by score.
Screenshot 2025-01-17 at 3 13 45 PM

3. Clear the search term in the Search Box. A re-fetch action should be triggered, and the results should once again be sorted by title.

### DIFF
--- a/src/components/search/helper/CleanHtml.tsx
+++ b/src/components/search/helper/CleanHtml.tsx
@@ -1,5 +1,0 @@
-export const stripHtmlTags = (html: string | null | undefined) => {
-  if (!html) return "";
-  const doc = new DOMParser().parseFromString(html, "text/html");
-  return doc.body.textContent || "";
-};
--- a/src/store/slices/searchSlice.ts
+++ b/src/store/slices/searchSlice.ts
@@ -2,7 +2,6 @@ import { createSlice, createAsyncThunk } from "@reduxjs/toolkit";
 import SolrQueryBuilder from "../../components/search/helper/SolrQueryBuilder";
 import { generateSolrObjectList } from "meta/helper/solrObjects";
 import { initialState, SolrSuggestResponse } from "@/store/types/search";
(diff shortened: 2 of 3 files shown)


mradamcox (changes requested):
Great, tested and works as expected. What is interesting to note about this is that with `score` as the sort, solr was placing the most recently updated record at the bottom of the list (that's how I realized that alphabetical wasn't being used). Anyway, it's good now.

One thing: Could you also remove the "CleanHtml" file? @bodom0015 and I noticed that was unused yesterday and you removed the import here but we should clean the file out too. Thank you!

pengyin-shan:
> Great, tested and works as expected. What is interesting to note about this is that with `score` as the sort, solr was placing the most recently updated record at the bottom of the list (that's how I realized that alphabetical wasn't being used). Anyway, it's good now.
> 
> One thing: Could you also remove the "CleanHtml" file? @bodom0015 and I noticed that was unused yesterday and you removed the import here but we should clean the file out too. Thank you!

Ah yes, I removed the file as well—thank you for the reminder. In Solr, when a matchAllDocs query is executed without additional sorting configurations, Solr defaults to sorting by the order in which documents were indexed so the new doc will be at the bottom I think, and Solr deals with 'MatchAll' as a special case I believe, not really following regular defined rules. By adding a MatchAll setting (see https://github.com/healthyregions/SDOHPlace-MetadataManager/pull/39), we can handle this issue on both ends, ensuring more predictable query results.

PR #420, opened 2025-01-23
[WIP] Security add

pengyin-shan:
Test for security setting with environmental variables


pengyin-shan:
Won't need this PR anymore since configure are done from the Solr side

PR #434, opened 2025-01-30
AI Test Search and Suggest Update

pengyin-shan:
This PR is for #434 and #439. It updates the suggestion scoring system to include more results and introduces the trial version of the LLM-based search experience. Additionally, it incorporates layout adjustments following the latest design changes and fixes several minor bugs.

## How to Test
1. Navigate to the search page. The keyword search should be the default mode:
Screenshot 2025-02-07 at 10 09 44 AM

2. Test the suggestion dropdown behavior: Start typing in the keyword search field. The suggestion dropdown should now correctly overlap the Geo search box on the map, fixing a previous issue: Screenshot 2025-02-07 at 10 09 56 AM

3. Verify keyword search functionality: Perform various searches to confirm that keyword search still works as expected: Screenshot 2025-02-07 at 12 23 28 PM

4. Switch to AI-based search mode: Click the "Chat" button to switch to AI search: Screenshot 2025-02-07 at 12 23 43 PM

5. Test AI-based search: Type a question (e.g., "What is the greenspace condition in Champaign, IL?") and search.
While searching, the mode switch, search, and clear buttons should be temporarily disabled to prevent conflicts caused by multiple triggered searches: Screenshot 2025-02-07 at 12 28 46 PM

6. Review AI search results: You should now see the AI-generated results: Screenshot 2025-02-07 at 12 27 24 PM

7. Switch back to keyword search: You can return to keyword search mode and test functionality as needed.


mradamcox:
@pengyin-shan based on the latest demo, have some thoughts about the map integration. To me, the ideal solution to the location/bbox search is not to have ChatGPT generate (as you currently have it), but instead to have the ChatGPT results trigger a selection in the [geocoding widget](https://docs.maptiler.com/sdk-js/modules/geocoding/api/api-reference/#geocoding#setQuery), assuming that is possible. That would take care of selecting the place in the map, zooming to it, and adding the bbox filter, instead of short-circuiting straight to the solr query, like the current setup. Do you think that is something you could explore before we merge this? Just wanted to put the idea out there.

PR #437, opened 2025-02-04
style: adjust styling for result cards within the Discovery App

bodom0015:
## Problem
Our designs for Result Cards within the Discovery App have changed

## Approach
* style: adjust styling / layout to match new designs

Before:
Screenshot 2025-02-04 at 4 58 37 PM

After:
Screenshot 2025-02-04 at 4 58 08 PM

## How to Test
1. Navigate to https://deploy-preview-437--cheerful-treacle-913a24.netlify.app/search
    * You should see result cards listed on the left side
    * You should see that the layout / styling has been adjusted to match [the latest designs](https://www.figma.com/design/vD6pC3DsFtGG2aCv4vWXtV/HEROP?node-id=3099-1201&t=Pf64kLoNiSa6xio2-0)

### DIFF
--- a/src/components/search/discoveryArea.tsx
+++ b/src/components/search/discoveryArea.tsx
@@ -2,12 +2,13 @@
 import { useEffect, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import type { AppDispatch, RootState } from "../../store";
-import { Grid } from "@mui/material";
+import {Collapse, Grid} from "@mui/material";
 import SearchArea from "./searchArea";
 import DetailPanel from "./detailPanel";
 import { setSchema } from "@/store/slices/searchSlice";
 import MapPanel from "./mapPanel/mapPanelContent";
 import dynamic from "next/dynamic";
+import * as React from "react";
 
 const DynamicResultsPanel = dynamic(() => import("./resultsPanel"), {
   ssr: false,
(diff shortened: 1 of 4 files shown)


mradamcox:
> For 5 and the additional comment at the end, this PR was focusing on the result cards themselves. But I can certainly try to take a look at the 6/6 split for the top purple bar + possible options for slide animations as well :)

Thanks!

bodom0015:
@mradamcox I think I've addressed all of these :+1: the 6/6 split should now be reflected in the top bar as well as the main results area (which was previously 5/7 instead of 6/6)

I see an animation when the panel expands but I'm not seeing an animation when the panel collapses again, which is odd.. I think we saw this previously as well (possibly with ChiVes?)

mradamcox:
> I see an animation when the panel expands but I'm not seeing an animation when the panel collapses again, which is odd.. I think we saw this previously as well (possibly with ChiVes?)

I see this too (firefox). What do you propose? We want these transitions to match of course.

mradamcox:
@bodom0015 Did you try the `easing` prop? https://mui.com/material-ui/api/collapse/

bodom0015:
I can definitely try playing around with the `easing`, and hopefully that fixes the timing?
I see what maybe could be happening here..

When user clicks "Get started", the panel contents change to become larger - this increases the max height of our SearchArea component.

When the user clicks close, the contents shrink immediately which also shrinks the max height of our div almost immediately. I suspect that the transition doesn't have time to change gradually, since the div changes sizes immediately.

Something else might be needed here to either 1) make the div height static (and not determined by contents, which seem wrong) or 2) slow down the speed that the contents change to provide the transition more time to react

This is getting complicated 😅 how high of a priority is the animation bit here?

mradamcox:
Well, if `easing` properties are that's needed, then this will easy to complete. Otherwise, if continues to be more complex, as you describe, let's just merge as is with the slide to expand transition, and then we won't be getting _too_ far away from the original point of this PR. Then we can come back in a different one to improve the transitions. Thanks for looking into this!

mradamcox (approved):
Ok, the updates look good now. We'll handle the transition later.

PR #446, opened 2025-02-11
Help bar

mradamcox:
Some style and content updates to the top help panel and related components:

- Move the panel to the left side, "behind" the page title text
- Add transition slide to info panel
- Add transition slide to filter panel
- Update text below the map to be more relevant, not overlap content in info panel


pengyin-shan (changes requested):
Looks good, just a few sugestions:
1. Update the terminology to align with the tooltip by using "AI-Inspired Search" instead of "Chat Search" since it is the one that we are using on mode-switch, and the current implementation is not conversational but instead prompts users to enter a single question.
Screenshot 2025-02-12 at 3 37 21 PM

2. Revise the current _The model "thinks" and returns to us a constructed set of query parameters that we then send to our search engine, Solr, and the response from Solr updates the search results. From the LLM we also get the "thought process" behind its response, which will appear below the search box._ with `LLM analyzes queries through semantic mapping and term relationship scoring to generate optimized Solr search parameters and retrieve relevant results from our database. It applies defined scoring algorithms and SDOH domain knowledge to enhance search accuracy. The analytical process, particularly query construction logic, is displayed below the search box` to make the "thinking" process clearer and eliminate potential doubts about the LLM's function.

3. Add a short message at the end with a link to collect user feedback: `Let us know your thoughts about this feature: link_to_contact_page)`.

Issue #456, opened 2025-02-19
[BUG] Disable the search when ai mode is enabled in default


pengyin-shan:
If a user enters the page with AI mode enabled but without a defined query (e.g., by pasting a URL like `/search?query=*&ai_search=true` into the browser), instead of allowing the AI to perform a search with an empty query, the user should be directed to an "All Results" page, where results are queried directly from Solr.

Issue #454, opened 2025-02-18
[BUG] Block the search action when ai-search is enabled and there's nothing in the search box


pengyin-shan:
We need to prevent AI-search submissions when the search box is empty. Unlike keyword-based search, AI-driven search triggers a "selection" behavior suggested by ChatGPT, assuming the user intends to view only SDOH data. As a result, not all data will be returned.

Issue #452, opened 2025-02-18
Change OpenAPI to UIUC.chat


pengyin-shan:
For security and efficiency reasons, change the OpenAPI to NCSA-Hosted LLM (UIUC.chat) for AI-based search

pengyin-shan:
As a note, this issue has been addressed in PR #457. The app has been configured to use the direct OpenAI API, uiuc.chat with OpenAI, and uiuc.chat with an NCSA-hosted open-source model by modifying environment variables, primarily LLM_ENDPOINT and MODEL_NAME.

As of March 2025, after discussions with the uiuc.chat team, we discovered that the current open-source models (specifically the NCSA-hosted LLMs) are unable to generate structured answers like GPT.

Additionally, uiuc.chat is RAG-based and designed to generate responses based on user-defined contexts, such as PDFs or Excel files, rather than the broader context that GPT models utilize—enriched by the large datasets OpenAI feeds into its models. Unfortunately, open-source models, including Llama, DeepSeek, and Qwen, are unable to match this level of contextual understanding. I attempted to configure them on my end and consulted the uiuc.chat team for prompt optimization, but they still fell short.

For now, we will continue using direct OpenAI API calls, as they remain the fastest and most accurate option for our needs. I’ve spoken with the uiuc.chat team, and they’ll notify us if any models that meet our requirements become available.

Issue #445, opened 2025-02-11
Update README file to reflect the new deployment process


pengyin-shan:
Since the introduction of the Netlify Edge feature in LLM Integration search, add instructions to Netlify on how to run it locally.

PR #457, opened 2025-02-20
[WIP] Switch to uiucchat for ai-based search and bug fixes on searching process

pengyin-shan:
This PR targets the issue of #456, #454, #452, #444, #445.

### DIFF
--- a/netlify/edge-functions/chat-search.js
+++ b/netlify/edge-functions/chat-search.js
@@ -1,4 +1,4 @@
-import { message } from "../../config/prompt/prompt_message.js";
+import { message } from "../../config/prompt/prompt_message_backup.js";
 
 const corsHeaders = {
   "Access-Control-Allow-Origin": "*",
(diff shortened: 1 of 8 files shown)


netlify/edge-functions/chat-search.js
```diff
@@ -20,21 +21,31 @@ function handleOptions(request) {
...
+    const courseName = Deno.env.get("COURSE_NAME");
+    const modelName = Deno.env.get("MODEL_NAME");
+    if (
+      !openAPIKey ||
+      !uiucChatAPIKey ||
+      !llmEndpoint ||
+      !courseName ||
+      !modelName
+    ) {
       console.error("API key not found in Deno environment");
```
mradamcox, line 38: @pengyin-shan Is there no way to run this without an OpenAI key?

pengyin-shan: @mradamcox, ah yes, if we're using an NCSA-hosted LLM, which is an open-source model, then the OpenAI key won't be necessary. I'll update the code to allow flexibility for this.

However, please keep in mind that, for now, we need to use a GPT-based model for our app due to its strong contextual capabilities, which open-source LLMs currently lack. This means we must provide an OpenAI key (see my explanation in the ticket description). That said, I'll modify the code to ensure it works seamlessly with an NCSA-based LLM when the OpenAI key isn't required.

mradamcox: Ok great! Thanks. This all looks really good overall.

pengyin-shan: Please see my latest commit for the update.

By the way, here’s a demonstration of why we need to use GPT for now:

- GPT-4o-mini: Follows our prompt, exports JSON correctly, and successfully derives the context term for us, based on its strong context pre-trained by OpenAI:
Screenshot 2025-02-24 at 12 37 38 PM

- DeepSeek (one of the open-source models, I have tried them all and got similar problems): While impressive, it only exports plain text with excessive wording and relies on saved documents. Extracting useful information from it is much harder:
Screenshot 2025-02-24 at 12 39 43 PM

mradamcox (approved):
Great, looks good, thanks!

netlify/edge-functions/chat-search.js
mradamcox: Ok, interesting, I see what you mean. Good thing to keep in mind for future integration then.

PR #465, opened 2025-02-26
fix: remove "shadow element" that appears when Overlays dropdown is open

### LINKED ISSUE #462: Fix ghost element when resource dropdown menu is active
I hacked this dropdown menu together a while ago, and just noticed that whenever it is open, there is a strange "ghost" element/box-shadow visible on the right hand of the screen. This needs to be removed.

Image

The style and look of the overlays dropdown is basically in line with the style of other popups, but it certainly could be changed in any way. The only important thing that must remain is that the labels of active overlays are shown (as now) when the dropdown is closed, and that the checkbox in the dropdown allows you to select multiple overlays.

bodom0015:
## Problem
On the Discovery App (Search view), when the Overlays dropdown is open we see a mysterious shadow element appear all the way down the right side of the app

We want to hide/remove this element when the dropdown is open

Fixes #462 

## Approach
* fix: remove the shadow element that appears when the Overlays dropdown is open

## How to Test
1. Navigate to https://deploy-preview-465--cheerful-treacle-913a24.netlify.app/search
2. Click the Overlays dropdown on the top-right
    * You should see that the dropdown opens to show the Overlays that we offer
    * You should not see a shadow / extra box appear over the scroll bar on the right
    * Compare with https://discovery_app.dev.sdohplace.org/search?query=*

### DIFF
--- a/src/components/search/mapPanel/mapPanelContent.tsx
+++ b/src/components/search/mapPanel/mapPanelContent.tsx
@@ -155,6 +155,7 @@ const MapPanelContent = (props: Props): JSX.Element => {
             id="basic-menu"
             className="flex items-center sm:justify-end mt-0 order-1 sm:order-none flex-none text-l-500 sm:mr-[2.3em]"
             style={{
+              margin: 0,
               boxShadow: "#aaaaaa 6px 12px 16px -8px",
             }}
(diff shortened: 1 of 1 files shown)


pengyin-shan (approved):
Tested and looks good to me. Approving this PR

Issue #470, opened 2025-03-03
[BUG] The Share function is not working, focusing on 'show-detail' function


pengyin-shan:
Currently, the 'Share' function is not working as expected. When a user clicks the shared link, the copy function works correctly. However, after pasting the link into the browser, the detailed view does not appear. The results did get filtered, except for the geolocation, which will be taken care of by #441 (plan to implement after soft release in mid-March).

PR #473, opened 2025-03-03
#470: Fix show detail

pengyin-shan:
This PR fixes issue #470, resolving a bug that prevented the detail view from appearing when users copied and pasted a URL containing the `show=` parameter.

## How to Test
1. Navigate to the search page and click "Details" on any result item to open the detail view.
2. Click "Share" to generate a shared link, then copy and paste the URL into a new tab or window.
3. The detail view should be displayed by default on the newly opened page.


mradamcox (approved):
Tested, works as described. Thanks!

Issue #471, opened 2025-03-03
[BUG] Add more space between the expanded filter and the first result item


pengyin-shan:
We need more space between the bottom of the expanded filter and the first result card:

<img width="729" alt="Image" src="https://github.com/user-attachments/assets/fe914736-bb32-4eac-aebd-7b8fbc59ceea" />

PR #474, opened 2025-03-04
#471: Add more bottom margin to the filter panel when it is on

pengyin-shan:
## PR for #471: Add Bottom Margin to Expandable Filter Panel
This PR introduces a bottom margin between the expandable filter panel and the first result item, improving visual separation. Previously, the panel and the first result appeared too close together:

Screenshot 2025-03-04 at 9 43 37 AM

## How to Test
1. Navigate to the search page and expand the filter panel.
2. You should now see a visible gap between the bottom of the panel and the first result item:
Screenshot 2025-03-04 at 9 43 18 AM

### DIFF
--- a/src/components/search/resultsPanel/index.tsx
+++ b/src/components/search/resultsPanel/index.tsx
@@ -18,8 +18,8 @@ import {
   resetFilters,
 } from "@/middleware/filterHelper";
 import ThemeIcons from "../helper/themeIcons";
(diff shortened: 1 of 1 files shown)


mradamcox (approved):
Perfect, thanks for this. I believe this was a regression from when I removed the top margin of the search results list, in order to line that up with the top of the map. I didn't test/realize it would affect the spacing of the filter panel when that is active.

PR #475, opened 2025-03-04
fix: null object pattern - use empty array if tags are undefined/null

bodom0015:
## Problem
This problem was located in #461 

When creating a new post, if `tags` are not defined in the frontmatter then the build/publish step can fail

For example:
https://github.com/healthyregions/SDOHPlace/pull/461/files#diff-0a533058926c6579ba4e5f979149f98902edf0d0aab98da670a6d891c2b96256R1-R10

## Approach
* fix: fallback to empty array if undefined/null tags are encountered

## How to Test
1. Create a post without specifying any `tags`
2. Click Publish and verify that the build completes successfully

### DIFF
--- a/src/pages/news/[post].tsx
+++ b/src/pages/news/[post].tsx
@@ -84,7 +84,7 @@ export const getStaticProps: GetStaticProps = async ({ params }) => {
       dateString: data.date,
       slug: slug,
       description: "",
-      tags: data.tags,
+      tags: data.tags || [],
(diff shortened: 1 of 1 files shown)


mradamcox (approved):
Awesome, thanks for figuring this out. I feel like Decap should be handling this better in the form validation (because we do have this marked as a required field) but this is definitely the best way for us to handle it now.

Issue #482, opened 2025-03-05
[BUG] Adjust sort options


pengyin-shan:
A behavioral issue that may be related to the filter condition adjustment ticket is that the sort options sometimes do not correctly reflect the expected sort condition. Check and fix the problem if it exists.

Issue #481, opened 2025-03-05
[BUG] Check all filter operations to make sure they work as expected after being shared


pengyin-shan:
Some filters are not performing the filtering operation after users use the Share function to allow a page re-render with pre-set filters. Check all filters to ensure they work as expected.

As a side note, the Sort and Geo-based search (filters triggered from the map) are not currently implemented with this feature. These can be added after the soft launch.

PR #487, opened 2025-03-06
#472,#481,#482, #488: adjust filter mechanism to avoid possible racing condition and minor bug fixes

### LINKED ISSUE #472: [BUG] Occasional overwriting of filtered results occurs after pasting the URL directly into the browser.
Occasionally, when pasting a filtered URL into the browser (e.g., ?query=*&subject=Employment&index_year=1963-2020&show=herop-zemals), the filtered results may unexpectedly be replaced by the 'return all' results.Users may see a flash of the correct results before they are replaced by an "All data sources" result.

pengyin-shan:
This PR fixes #472, #481, #482 and #488. It addresses occasional race conditions that occur when triggering filters, as well as an occasional double-trigger of data fetching during initialization. To resolve these issues, I have implemented a request ID mechanism and adjusted the trigger points for fetch activities, as well as re-organized reset filter methods. These changes should help prevent unintended behavior.

## How to Test
1. Navigate to the search page and test the three sort options. They should function correctly, with the appropriate underline highlighting the current sorting option for #482:
Screenshot 2025-03-06 at 3 48 14 PM

*(Compare this to the current behavior at https://discovery_app.dev.sdohplace.org/search)

2. Perform various filtering actions to verify they work as expected. Then, copy the URL (or use the 'Share' function), paste it into a new tab, and ensure the new page reflects the same state.
Note: The map does not yet sync with the URL, so it always loads in its initial state. Additionally, sorting does not persist via URL, so the default sorting order remains 'Relavence'.

3. Repeat step 2 multiple times to confirm that issues #472 and #481 are resolved.

4. #488 Fix: Start a fresh search, then click "Details" for **10-minute walk service areas (10MWSA)** to open the detail panel. Next, apply a filter by theme: "Demographics."

Screenshot 2025-03-07 at 12 18 59 PM

You should notice that the map view reappears because **10-minute walk service areas (10MWSA)** is not included in the filtered results. For comparison, check https://discovery_app.dev.sdohplace.org/search?show=herop-jxbloa&subject=Demographics&query=* to see the difference.

### DIFF
--- a/src/components/map/dynamicMap.tsx
+++ b/src/components/map/dynamicMap.tsx
@@ -31,6 +31,7 @@ import GeoSearchControl from "./geoSearchControl";
 import { clearMapPreview } from "@/store/slices/uiSlice";
 import {EventType} from "@/lib/event";
 import {usePlausible} from "next-plausible";
+import { debounce } from "@mui/material";
 
 const apiKey = process.env.NEXT_PUBLIC_MAPTILER_API_KEY;
 
@@ -201,7 +202,7 @@ export default function DynamicMap(props: Props): JSX.Element {
   }, []);
 
   const setBboxOnMoveEnd = useCallback(
-    (event: ViewStateChangeEvent) => {
+     debounce((event: ViewStateChangeEvent) => {
       const bounds = event.target.getBounds();
       const newBbox: [number, number, number, number] = [
         Math.round(bounds._sw.lng * 1000) / 1000,
@@ -210,8 +211,7 @@ export default function DynamicMap(props: Props): JSX.Element {
         Math.round(bounds._ne.lat * 1000) / 1000,
       ];
       dispatch(setBbox(newBbox));
-    },
-    [dispatch]
+    }, 300),[dispatch]
   );
 
   const handleGeoSearchSelection = useCallback(
--- a/src/components/search/discoveryArea.tsx
+++ b/src/components/search/discoveryArea.tsx
@@ -2,7 +2,7 @@
 import { useEffect, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import type { AppDispatch, RootState } from "../../store";
-import { Collapse, Grid } from "@mui/material";
+import { Grid } from "@mui/material";
 import SearchArea from "./searchArea";
 import DetailPanel from "./detailPanel";
 import { initializeSearch, setSchema } from "@/store/slices/searchSlice";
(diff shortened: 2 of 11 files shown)


mradamcox (changes requested):
@pengyin-shan Looking at the sorting, I'm not getting the expected results. If I load the page and sort by "Oldest first" I get this item first: 
image
even though I know there is an item with a date range that goes back to 1963. Similarly, sorting by latest puts non-2024 items at the top, even though there are items from 2024.

Points 2) & 3) look good. For 4), see my comment on the underlying ticket regarding a different way to handle that behavior.

pengyin-shan:
> @pengyin-shan Looking at the sorting, I'm not getting the expected results. If I load the page and sort by "Oldest first" I get this item first, even though I know there is an item with a date range that goes back to 1963. image Similarly, sorting by "Recent first" puts non-2024 items at the top, even though there are items from 2024.
> 
> Points 2) & 3) look good. For 4), see my comment on the underlying ticket regarding a different way to handle that behavior.

Hi @mradamcox , to confirm, the sort field I used for most recent and oldest is `date_issued` instead of `index_year`. It seems that we are not on the same page?

pengyin-shan:
> > @pengyin-shan Looking at the sorting, I'm not getting the expected results. If I load the page and sort by "Oldest first" I get this item first, even though I know there is an item with a date range that goes back to 1963. image Similarly, sorting by "Recent first" puts non-2024 items at the top, even though there are items from 2024.
> > Points 2) & 3) look good. For 4), see my comment on the underlying ticket regarding a different way to handle that behavior.
> 
> Hi @mradamcox , to confirm, the sort field I used for most recent and oldest is `date_issued` instead of `index_year`. It seems that we are not on the same page?

I was thinking the 'recent' means the most recent data we have added, and index_year will be handled by the filter

mradamcox:
> I was thinking the 'recent' means the most recent data we have added, and index_year will be handled by the filter

Ah, that totally makes sense, but I think we should use index year for the sorting as well, as in "Oldest first" gives us the oldest dataset in that matches the current search results.

pengyin-shan:
> > I was thinking the 'recent' means the most recent data we have added, and index_year will be handled by the filter
> 
> Ah, that totally makes sense, but I think we should use index year for the sorting as well, as in "Oldest first" gives us the oldest dataset in that matches the current search results.

Yeah, I think we can sort by index year if we want. For an item with multiple index years, which one should we use for comparison—the first year or the last year?

mradamcox:
For "Oldest first" use the earliest index year, for "Recent first" use the latest index year.

mradamcox:
Ugh, this merge conflict is my fault, these were auto imports that VSCode put in by mistake but I didn't want and shouldn't have committed:
image
Whoever resolves this, that `ConstructionOutline, Layers` import line can be fully removed.

pengyin-shan:
> Ugh, this merge conflict is my fault, these were auto imports that VSCode put in by mistake but I didn't want and shouldn't have committed: image Whoever resolves this, that `ConstructionOutline, Layers` import line can be fully removed.

Awesome, I'll keep this in mind

pengyin-shan:
Hi @mradamcox , the sort should be adjusted by the `index_year` now, while items w/ out index_year always showing at the very bottom. This behavior should be fixed once we add index_year to all data:
Screenshot 2025-03-12 at 10 54 48 AM

Issue #496, opened 2025-03-12
Update scrollbar to be browser/platform agnostic


shubhamk008:
Here is the CSS we could use for the containers that need scrolling. Let me know once done, might need to tweak the colors a bit:

```
body ::-webkit-scrollbar {
    width: 5px;
    height: 5px
}

body ::-webkit-scrollbar-button {
    width: 0;
    height: 0
}

body ::-webkit-scrollbar-thumb {
    background: #dad8d8;
    border: 0 none #fff;
    border-radius: 0
}

body ::-webkit-scrollbar-thumb:hover {
    background: #dad8d8
}

body ::-webkit-scrollbar-thumb:active {
    background: #000
}

body ::-webkit-scrollbar-track {
    background: #f4f6f9;
    border: 0 none #fff;
    border-radius: 0
}

body ::-webkit-scrollbar-track:hover {
    background: #f4f6f9
}

body ::-webkit-scrollbar-track:active {
    background: #333
}

body ::-webkit-scrollbar-corner {
    background: 0 0
}
```

mradamcox:
One point of clarification: What is the scope of where this should be applied? Should this be implemented across the entire website (blog posts, showcase pages, home page, discovery app, etc.), or only within the discovery app, or only on certain elements within the discovery app? There's a good chance we'll be able to make progress on it before the soft launch.

pengyin-shan:
I'll handle this task. @shubhamk008, just to confirm—we only want the scrollbar for the result list to be always visible, correct?

shubhamk008:
Yes, I would say let us leave the other places for now and only implement it for Data Discovery. So the results section and the Get started help box

PR #506, opened 2025-03-13
#496 agnostic scrollbar

pengyin-shan:
This PR addresses issue #496. It applies the styling approach introduced by @shubhamk008 to add a permanent scrollbar to the result list and the help-information section. I have tested it on Chrome, Edge, and Safari on my MacBook.

Note that the WebKit-based solution does not work for Firefox (https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar). While a non-WebKit method can alter the scrollbar’s appearance (like https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scrollbars_styling), it does not ensure a permanently visible scrollbar. Instead, users can enable this feature manually in Firefox settings (https://connect.mozilla.org/t5/discussions/scrollbars-disappear-in-firefox-100/m-p/6358). If I missed any relevant resources, please let me know—I did some searches but couldn't find a proper solution.

However, when users hover over a scrollable section, a scrollbar briefly appears and disappears, indicating that scrolling is available. Hopefully, this helps improve usability.

## How to Test
1. Navigate to the search page. A scrollbar should be visible in the result list after the initial render.
Screenshot 2025-03-13 at 12 49 33 PM

2. Check the help text section. The scrollbar should be present for the vertically scrollable info panel.
Screenshot 2025-03-13 at 12 49 39 PM

3. Test across multiple browsers to verify the behavior.

### DIFF
--- a/src/components/search/resultsPanel/index.tsx
+++ b/src/components/search/resultsPanel/index.tsx
@@ -159,7 +159,7 @@ const ResultsPanel = (props: Props): JSX.Element => {
               {isLoading || isResetting || isInitialLoad ? (
                 renderLoadingState()
               ) : (
-                <div>
+                <div className="force-scrollbar">
                   {searchState.results.length > 0 ? (
                     <Box
                       height="100%"
--- a/src/components/search/searchArea/infoPanel.tsx
+++ b/src/components/search/searchArea/infoPanel.tsx
@@ -163,6 +163,7 @@ export default function InfoPanel() {
         </Tabs>
       </Box>
       <Box
+        className="force-scrollbar"
         sx={{
           transition: "left 0.3s linear",
         }}
--- a/src/styles/globals.css
+++ b/src/styles/globals.css
@@ -297,3 +297,46 @@ article table tbody tr:last-child {
 article table td {
   text-align: left;
 }
+
+/* Scrollbar */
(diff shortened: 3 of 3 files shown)


mradamcox (approved):
Great, this looks good to me! FWIW, the default scrollbars on Firefox look more similar to the desired design anyway, than the defaults on other browsers. But you're right, they're appearance isn't forced. To me, not a huge deal, FF is such a small share of the browser market.

image

shubhamk008:
Can we change the black and gray in the video to mimick the default colors (gray and light gray) for the scrollbar?
[Drive](https://drive.google.com/file/d/15cvZgI4MH4G8mmO2A3J6GvzlbdsjE2Mx/view?usp=sharing)

pengyin-shan:
> Can we change the black and gray in the video to mimick the default colors (gray and light gray) for the scrollbar? [Drive](https://drive.google.com/file/d/15cvZgI4MH4G8mmO2A3J6GvzlbdsjE2Mx/view?usp=sharing)

Hi @shubhamk008 , my latest update should fix this issue. Let me know if you still see the problem

mradamcox:
I'm not seeing the dark background now in Chrome, so I'm going to go ahead and merge this in. Looks good!

PR #508, opened 2025-03-13
feat: rework advisory page and CMS

### LINKED ISSUE #495: Changes to Advisory page and CMS
Currently the Advisory page has two sections for the two different advisory boards we have had in the past. Going forward we need this to be changes into the following two sections: `Current Advisory Board` and `Past Advisory Board Members`, and, as now, the entries in each section will be populated by the Decap CMS content. However, changes to that content will be necessary...

Currently there are two separate schemas in the CMS for "Technical Advisory Group" and "Stakeholder Advisory Group":
Image

What should happen is a new schema is made alongside these just called "Advisory", and all entries from the existing two groups are combined into that new group. Then, add a new field for "Status" with options "current" and "past", and sort all person entries with this field. Once the frontend is configured properly, it will populate the new Advisory page with these sections.

Then, the old "Technical Advisory Group" and "Stakeholder Advisory Group" schemas should be removed from the CMS.

bodom0015:
## Problem
We would like to distinguish between past and current Advisory Board members

Fixes #495 

## Approach
* feat: merge `stakeholder_advisory.json` & `technical_advisory.json` into a new `advisory.json` in DecapCMS backend
* feat: add field for `Status` for Advisory Board members
* feat: update Advisory page in the frontend to reflect Current/Past Advisory Board members

Screenshot_2025-03-13_at_11_47_24_AM

Screenshot 2025-03-13 at 11 46 30 AM

Screenshot 2025-03-13 at 11 49 49 AM

## How to Test
1. Navigate to /admin/index.html
2. View the People -> Advisory Board section
    * You should see 17 people have been added here - this should include all people from stakeholder & tech advisory
    * You should see that the Current & Past Advisory Board members are listed here (based on Status)


mradamcox:
@bodom0015 Splendid, all of the code looks perfect here. I'm not sure it's a problem, but if I go to the Decap backend on the deploy preview `/admin` then do see the new collection but not the entries. I believe this is because the deployed Decap backend looks directly at the publish branch, but just wanted to check with you that that sounds right.

bodom0015:
I think that sounds right? 🤔 Hopefully the entries appear when we merge `main` back into `publish`, but if not I can take a closer look then :+1:

Issue #507, opened 2025-03-13
[BUG] Lost click event on map preview button in result item list


mradamcox:
## problem

If I have clicked the "Show on map" button for an item in the result list, the map preview appears like this (good):

![Image](https://github.com/user-attachments/assets/caeeb24a-b3cb-4b59-a003-6af9d5673065)

However, if I click on the "Show on map" button in the item below one time, nothing happens. If I click it a second time, then, as expected, the preview is activated on the second layer and removed for the first layer. No errors appear in the dev console.

![Image](https://github.com/user-attachments/assets/3a4b6c92-dad2-43a8-bfcb-2840a5036582)

Observed on Chrome and Firefox.

## expected behavior

Clicking on any "Show on map" button will immediately activate that layer's preview on the map and deactivate any existing preview.

pengyin-shan:
This issue stems from the limited clickable area of the 'Show on Map' button, which can lead to accidental clicks. I addressed this in #510 by reintroducing the checkbox and adding some additional visual effects.

![Image](https://github.com/user-attachments/assets/7071f070-4589-4075-99a5-9492bd341097)

@mradamcox  @shubhamk008, let me know if you prefer the previous version, and I'll revert the changes. The concern with the current label-only version in production is that it's harder to tell if the button has been clicked and provides less clickable space for users.

In the future, we could further improve the UI to better distinguish the 'Show on Map' action from the default result card click behavior.

PR #510, opened 2025-03-13
#507: Improve the click experience for "Show On Map" in result card

pengyin-shan:
This PR addresses issue #507. It aims to resolve the problem where the 'Show On Map' link in the result card is too small, making it prone to mis-clicks since the surrounding area is also clickable.

To improve the click experience, I’ve introduced some UI changes, including adding a checkbox to indicate the button’s effect back. If the previous version is preferred, please let me know, and I’ll revert the changes.

## How to Test
1. Visit https://deploy-preview-510--cheerful-treacle-913a24.netlify.app/search and hover over a 'Show on Map' button. A shadow should appear to indicate the clickable area:
Screenshot 2025-03-13 at 2 57 04 PM

2. Click the 'Show on Map' button. The map should update, and a checkmark should appear to confirm the button's effect:
Screenshot 2025-03-13 at 2 58 23 PM

3. Click another 'Show on Map' button. The map should update again, and the button's status should reflect the new selection:
Screenshot 2025-03-13 at 2 58 44 PM

4. Click the 'Remove Preview' button. The map overlay should be removed, confirming the click action.

To further improve the user experience, we may want to explore design changes such as increasing the button's font size to reduce accidental clicks between the result card (which triggers the detail view) and the 'Show on Map' button (which updates the map) in the future.

### DIFF
--- a/src/components/search/resultsPanel/resultCard.tsx
+++ b/src/components/search/resultsPanel/resultCard.tsx
@@ -1,7 +1,7 @@
 "use client";
 import { makeStyles } from "@mui/styles";
 import * as React from "react";
-import { Checkbox, FormControlLabel, Grid, Typography } from "@mui/material";
+import { Grid } from "@mui/material";
 import tailwindConfig from "../../../../tailwind.config";
 import resolveConfig from "tailwindcss/resolveConfig";
 import IconText from "../iconText";
@@ -341,7 +341,7 @@ const ResultCard = (props: Props): JSX.Element => {
         container
         className="flex flex-col sm:flex-row px-2 mt-1"
       >
-        <Grid item xs={8}>
+        <Grid item xs={8} sx={{ mt: "1em", pt: "0 !important" }}>
           <div className={`${classes.resultCard} truncate `}>
             Keywords:{" "}
             {props.resultItem.meta.keyword
@@ -355,7 +355,7 @@ const ResultCard = (props: Props): JSX.Element => {
               : ""}
           </div>
         </Grid>
-        <Grid item xs={4}>
+        <Grid item xs={4} sx={{ mt: "1em", pt: "0 !important" }}>
           <div className={`${classes.resultCard} truncate `}>
             Year:{" "}
             {props.resultItem.index_year?.length > 1
--- a/src/components/search/resultsPanel/resultCard.tsx
+++ b/src/components/search/resultsPanel/resultCard.tsx
@@ -1,7 +1,7 @@
 "use client";
 import { makeStyles } from "@mui/styles";
 import * as React from "react";
-import {Checkbox, FormControlLabel, Grid, Typography} from "@mui/material";
+import { Checkbox, FormControlLabel, Grid, Typography } from "@mui/material";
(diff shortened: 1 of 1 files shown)


mradamcox (changes requested):
@pengyin-shan I did originally have a checkbox when I first introduced this feature, but @shubhamk008 preferred without checkbox in the final designs. It seems like the changes you've made would still be big improvements even without the actual checkbox, right? If so, could you try removing the checkbox itself but leave all the other button behavior/size the same?

pengyin-shan:
I 

> @pengyin-shan I did originally have a checkbox when I first introduced feature, but @shubhamk008 preferred without checkbox in the final designs. It seems like the changes you've made would still be big improvements even without the actual checkbox, right? If so, could you try removing the checkbox itself but leave all the other button behavior/size the same?

During testing, I found that the checkbox version provides a better experience than the non-checkbox version. The checkbox not only visually indicates the result of the click action (e.g., when the map has a cover) but also enlarges the clickable area, making it easier to interact with.

Compared to the label-only version, the checkbox provides clearer feedback. If a checkbox is unchecked after a click, users are more likely to retry, whereas a label-only button might leave them uncertain about whether anything should have happened. Additionally, users can click either the checkbox or the label to achieve the same effect, reducing the likelihood of misclicks.

That said, I can certainly remove the checkbox and make the label the only clickable element 👍 . Just that I feel that the checkbox version is easier to use.

At the same time, we might also consider removing the clickable action from the entire card, allowing users to navigate to the detailed view only by clicking "Details →." This could further reduce the risk of accidental clicks.

mradamcox:
Sounds good, I agree with you--I prefer the checkbox and I don't think the entire card should be clickable. However, we'll wait for @shubhamk008's input before moving forward.

pengyin-shan:
> Sounds good, I agree with you--I prefer the checkbox and I don't think the entire card should be clickable. However, we'll wait for @shubhamk008's input before moving forward.

@mradamcox @shubhamk008 I’ve created a test PR (https://github.com/healthyregions/SDOHPlace/pull/511) for the non-checkbox version. Feel free to try it out since my personal experience is not representative, and let me know which version you prefer—I’ll use that as the final version.

mradamcox:
I think the underlying issue is the overlap of the element that is below, which makes a smaller-than-expected clickable area. It's a little tricky to show in a screenshot, but here is the size of what should be the clickable element
image
but the element below actually covers up the bottom portion of that area.
image

pengyin-shan:
> I think the underlying issue is the overlap of the element that is below, which makes a smaller-than-expected clickable area. It's a little tricky to show in a screenshot, but here is the size of what should be the clickable element image but the element below actually covers up the bottom portion of that area. image

Ah, good catch! I’ve changed the padding-top to margin-top to help handle this so the area from details below won't overlay

mradamcox:
Nice, that makes a big difference.

mradamcox:
Closing in favor of https://github.com/healthyregions/SDOHPlace/pull/511, thank you @pengyin-shan for working through this.

PR #512, opened 2025-03-13
feat: rework the fellows page to support multiple cohorts

### LINKED ISSUE #364: Facilitate multiple cohorts in fellows page
We have a new set of fellows to add to the CMS, and these will need to be listed in the fellows page, separate from the existing cohort. The CMS already has a category for Fall 2024 fellows, the frontend just needs to be updated to properly display all of the new fellows.

I think the way I'll approach this is

1. Create the code needed to show the section section, but
2. comment it out, until the CMS content has been added, then
3. uncomment when ready.

bodom0015:
## Problem
Our current Fellows page only displays the Spring 2024 cohorts

We would like to update this view to display and organize fellows from multiple different cohorts

Fixes #364 

## Approach
* fix: reword the Advisory page - Past Advisory Board Members -> Project Launch Advisory
* feat: rework the Fellows page to support multiple cohorts

Screenshot 2025-03-13 at 3 15 06 PM

Screenshot 2025-03-13 at 3 15 20 PM

Screenshot 2025-03-13 at 3 15 33 PM

## How to Test
1. Navigate to https://deploy-preview-512--cheerful-treacle-913a24.netlify.app/fellows
    * You should see the view now displays Spring 2024 Fellows
    * Once we add Fall 2024, you should see them displayed above Spring 2024 (since they are more recent)
    * Similarly, once we add the Spring 2025 Cohorts, they will appear above the 2024 ones

### DIFF
--- a/public/admin/config.yml
+++ b/public/admin/config.yml
@@ -123,7 +123,7 @@ collections:
               - label: "Cohort"
                 name: "cohort"
                 widget: "select"
-                options: ["Spring 2024", "Fall 2024"]
+                options: ["Spring 2024", "Fall 2024", "Spring 2025"]
               - label: "Links"
                 name: "links"
                 widget: "list"
--- a/public/admin/config.yml
+++ b/public/admin/config.yml
@@ -107,6 +107,7 @@ collections:
             label_singular: "Fellow"
             widget: list
             summary: "{{name}} - {{cohort}}"
+            add_to_top: true
             fields:
               - label: "Name"
                 name: "name"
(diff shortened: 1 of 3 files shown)


mradamcox (changes requested):
Clever approach, I like it! Could you just 1) remove the console log statements, and 2) add "Spring 2025" to the list of options in the Decap config for fellows?

mradamcox (approved):
Looks great to me, thanks! There's a chance we'll update the name of the past advisory board, but for now this is perfect.

Issue #517, opened 2025-03-14
[FEATURE] Add explaination to ai-help text about the disconnection between search behavior and the filter/map


pengyin-shan:
link to #441. Add some text to quickly explain that the map and filter UI isn’t currently connected to the AI search. Users will see location-filtered data though when their question has a location. The LLM reads location terms and translates them into a bounding box filter for the Solr query, but the map doesn’t have the capability to display this. So it feels a bit disconnected between the filter/map and the result.

Update example and address that this is something we will work on in the next stage

PR #518, opened 2025-03-14
#517: Update ai-search help text to explain current limitation

pengyin-shan:
This PR addresses issue #517. It clarifies that AI-Search results are filtered by location when specified in the user's query. However, this filtering is not yet visually reflected on the map or filter panel, which may confuse some users when they see filtered results without corresponding map updates.

## How to Test
1. Navigate to the search page and open the help panel for AI-Search.
2. An additional explanation should now appear at the end of the help text:
Screenshot 2025-03-14 at 11 41 01 AM

### DIFF
--- a/src/components/search/searchArea/helpText/aiSearchText.tsx
+++ b/src/components/search/searchArea/helpText/aiSearchText.tsx
@@ -106,15 +106,10 @@ export const AiSearchText = () => {
             fontSize: "0.8em",
           }}
         >
-          When you include location terms in your questions (like Chicago), our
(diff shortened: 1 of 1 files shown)


mradamcox (changes requested):
This is a good note, but I think it would be better to shorten a bit, how about:

While AI results will translate location names like "Chicago" into a geographic filter, this filter will not be visually depicted in the map interface. We're working toward better integration in our next update.

pengyin-shan:
@mradamcox , I updated the description to use the shorter version.
Screenshot 2025-03-14 at 2 10 03 PM

Issue #524, opened 2025-03-19
[BUG] Error when clicking the 'READ MORE' link for team bio with no link list


pengyin-shan:
The 'READ MORE' link on the team page (https://sdohplace.org/team) for Sara, Camrin, and Adrita are returning the following error:
<img width="1230" alt="Image" src="https://github.com/user-attachments/assets/9221d768-3224-4a2c-aa31-15325b67930d" />

## fix

The problem is that the front-end expects a list of one or more links to be attached to each person. This should be considered optional and the page should not break if there is no list of links provided.

mradamcox:
Looks to be because these entries have no links (like links out to personal website) associated with them. I'll update the title of this ticket accordingly, because the front end should handle not having links.

PR #528, opened 2025-03-19
fix: handle no link list in team person entry #524

mradamcox:
Tiny change, decided to just do it quickly. @pengyin-shan You can merge this in after review.

### DIFF
--- a/src/pages/team.tsx
+++ b/src/pages/team.tsx
@@ -95,7 +95,7 @@ const Team: NextPage = () => {
                   <div className="text-lg font-medium leading-[177.778%] mt-2.5">
                     {modalData.title}
                   </div>
-                  {modalData.links.map((link, index) => (
+                  {modalData.links?.map((link, index) => (
                     <div
                       key={index}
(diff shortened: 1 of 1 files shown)


mradamcox:
No, we shouldn't merge `publish` to `main`, but your point here is a good one. I think the way to test this would be to pull it locally and merge into a local branch made from publish, or something like that.

Overall, this is PR is a tiny change (and I did test locally with some modified JSON files that I didn't commit), but for more substantial changes to CMS content presentation we may need better mock data in the main branch.

One idea would be to add mock data to main, merge to publish, and then delete from publish. For something more extreme, we could even:

- save out all CMS content files from `publish`
- remove all existing CMS content from `main` and commit
- add new set of test/demo content to `main` and commit
- merge into `publish` (this will delete all production CMS files and add mock content)
- add backed up CMS content to `publish` and remove mock content, commit

I mean, that all feels a little hacky, but it would accomplish the underlying goal of only having demo content committed to the `main` branch, and only production CMS content committed to `publish`. Anyway, something to think about moving forward, but no need to address right now.

Issue #525, opened 2025-03-19
[bug] "no results" events are sent to Plausible but the payload isn't


mradamcox:
We have a custom event set up that will ping our analytics when someone runs a search but gets no results. We would like to have the payload for that custom event send along the search term that was used when "no results" was returned (it's likely this is mostly set up, already, just needs some small fix).

pengyin-shan:
Hi @mradamcox, just to clarify—are we looking for the term passed by the ReceivedNoSearchResults Plausible event? I selected the "Properties" option in the tab, which displays a list of terms. Based on the code, I believe these correspond to the search queries.

<img width="483" alt="Image" src="https://github.com/user-attachments/assets/a3ab4224-5b69-4367-ba1b-3bda5cddea68" /> 

<img width="847" alt="Image" src="https://github.com/user-attachments/assets/340b7dc1-6334-4f34-bf01-e0eeb327b41d" />

Looking at the queries, some should return results (e.g., "Food Environmental Index"), unless users are adjusting filters based on the query results. To gain more insight into search behavior, I’ll add a `searchFilters` props to track the applied filters and their values.

mradamcox:
Ah, good catch, you are totally right: Most of the hits for "No Search Results" are happening when multiple filters have been applied without an actual keyword search, so it makes sense there is no reported payload. I had only looked at these properties on yesterday's chart (when strangely there weren't any examples of keyword searches with no results), but as you show, in all time there have been many. Thanks!

Adding the search filter property sounds like a good addition that will help us understand this better.

PR #530, opened 2025-03-20
#525: Add filter status to Pausible payload to check for no result search

pengyin-shan:
This PR addresses issue #525 by adding filterStatus props to the payload passed to Pausible in the 'No Result' view.

As discussed in the ticket, most occurrences of "No Search Results" happen when multiple filters are applied, since the recommendation mechanism attempts to prevent 'No Result' views for single searches.

To enhance visibility into filter usage, this PR introduces two new props:

1. One represents filters as an array.
2. The other provides a more comprehensive string format.

Since it's unclear which format will perform better, both can be included initially. Once Pausible starts receiving results, we can remove any unnecessary props. With this change, we will be able to see the applied filters and their values when a user encounters a 'No Result' view.


mradamcox (approved):
Great, thanks for figuring this out and adding these properties. We'll just keep an eye on it over the next couple of weeks.

PR #532, opened 2025-03-20
feat: add Natural Environment subject icon and filter

### LINKED ISSUE #526: [Feat] New icon for "Physical Environment" subject
We have a set of icons already for all of the different subject filters. We need to add another one following the existing pattern, for a new subject called "Physical Environment".

@shubhamk008 will supply the icon.

mradamcox:
Closes #526.


mradamcox:
Tested this out after adding the option to the metadata manager schema, works fine:
image

Issue #529, opened 2025-03-20
[bug] Inconsistent titles on main pages


mradamcox:
The meta `<title>` tag needs to be updated in several places. The standard across the site should be

`<page name> | SDOH & Place Project` 

This is correct in most places, with a few notable exceptions that need to be fixed.

- Data Discovery doesn't have any title, it should be updated to `Data Discovery | SDOH & Place Project`
- The Core Team page has the incorrect name in title
- The Fellows Showcase page is titled `Showcase | News | SDOH & Place` but should not include "News" 
- The research guides pages (we have two so far) need to have a page name added to them (they are both just `SDOH & Place Project` right now)

Issue #514, opened 2025-03-13
[FEAUTE] Update styles


pengyin-shan:
1. Increase Spacing: Add more space between the icon and the start text for better visual balance.

![Image](https://github.com/user-attachments/assets/46357ff1-a673-445c-938c-19794edd8b14)

2. Future Improve Responsiveness: On smaller screens (10 inches), the right side layout feels cramped. Adjust the design to push the "Share" and "Go to" links to the next row for a cleaner, more responsive layout. 
![Image](https://github.com/user-attachments/assets/3a3002da-0435-44bc-878f-a1d25bd02e36)

3. Update Banner Color: Change the banner to a different color. (@shubhamk008, please refer to the Slack suggestion from @Makosak)

4. (We can discuss this after soft lanuch) Panel Color Update: Consider updating the panel color for better contrast or visual appeal.
![Image](https://github.com/user-attachments/assets/21b66b88-2ff4-4139-9a32-8fdb4b3d5ab5)

Makosak:
Consider a more pronounced border to the bottom of the left panel/search results. When exploring a dataset on the right side, the left part suddenly disappears.

<img width="1443" alt="Image" src="https://github.com/user-attachments/assets/e344ad8f-3612-4e21-8600-de463f42e9c1" />

Makosak:
After reviewing the data, we really need a new category to capture physical environment (as different from built environment) -- so need a new icon for "Physical Environment." Will capture ozone, air quality, EJscreen measures, etc.

mradamcox:
> After reviewing the data, we really need a new category to capture physical environment (as different from built environment) -- so need a new icon for "Physical Environment." Will capture ozone, air quality, EJscreen measures, etc.

I've created a separate ticket for tracking that task: https://github.com/healthyregions/SDOHPlace/issues/526

PR #533, opened 2025-03-20
Release 0.1.1

### LINKED ISSUE #524: [BUG] Error when clicking the 'READ MORE' link for team bio with no link list

mradamcox:
Trying out our new branch pattern for releases. This will merge all latest work from `develop` into `main`.

Closes #524, #525, #526, #529.

Also a addresses a few of the points in #514 (not all though).


pengyin-shan (approved):
This looks great—thanks, @mradamcox, for preparing this release!

Just an FYI: If we’d like, we can also include the version number in `package.json`, for example:
Screenshot 2025-03-20 at 5 21 20 PM
Or we can do `npm version 0.1.1` I believe to automatically add it, But this is just optional.

mradamcox:
> Just an FYI: If we’d like, we can also include the version number in package.json

Good idea, will do this next time.

mradamcox:
> Or we can do `npm version 0.1.1` I believe to automatically add it, But this is just optional.

This command actually creates a commit and tags it in git, so we should just update the number in the file itself directly, that way we leave all of the actual tag creation to the release notes process.

PR #536, opened 2025-03-21
#535: Add more information in Plausible to watch the search behavior

pengyin-shan:
This PR addresses issue #535 and should be included in the next micro release. It enhances search functionality by adding search-type information along with search values and filter details. This improvement will enable a more precise analysis of user search behavior.


mradamcox (approved):
Looks good, this will add a bit more context.

Issue #546, opened 2025-04-21
[BUG] Show the results with no direct result but has related result


pengyin-shan:
There’s a bug in the keyword search functionality. When a user searches for a term that doesn’t yield any direct matches (i.e., no auto-complete suggestions), but there are related results available through our recommendation mechanism, the system still displays a "No Results" view—despite showing a number of recommended results:

<img width="1126" alt="Image" src="https://github.com/user-attachments/assets/db60773b-0ee7-4290-9c12-c753fd9f8cb7" />

**Expected Behavior:**
Update the behavior so that when direct matches are unavailable but recommended results exist, users are presented with those related results instead of the "No Results" view. This will improve the user experience and better showcase relevant content.

Issue #542, opened 2025-04-11
[BUG] Pre-selected option in the suggestion list


pengyin-shan:
Prevent the behavior that if a selection had already been used, the second selection would not be stuck on the first item list

PR #548, opened 2025-04-21
#549, #542, #546: Fix the related term bug and update search box autocomplete behavior

pengyin-shan:
This PR addresses issues #549, #542, and #546. It updates the default behavior of the autocomplete dropdown in the search box to prevent pre-selection when the user's mouse moves out of the dropdown area. It also fixes some bugs related to the "No Results" scenarios and includes general code optimizations.

## How to Test

1. #542: Go to the search page and type a keyword longer than two characters to trigger the autocomplete dropdown (e.g., "scocial").

2. Hover over a suggestion, then move your mouse out of the dropdown.

3. Previously, the first item would be automatically highlighted (shado
Screenshot 2025-04-22 at 4 53 39 PM
wed), but now this is no longer the case.

4. Press Enter after moving your mouse out — the search should be performed using your typed input (e.g. "social"), not the first suggestion.
Screenshot 2025-04-22 at 4 54 05 PM

5. #546: Type a short term that doesn't trigger the dropdown (e.g., "sa"). Note that this term should not match any Solr suggestions but could still return related results.

6. Press Enter or click the arrow to search.

7. You should see a list of relevant results.
Screenshot 2025-04-22 at 4 54 22 PM

8. Previously, this scenario incorrectly showed a "No Results" view with only the result count visible.

9. #549: Type a term that yields no direct or related results (e.g., "sj") and press Enter.

10. The "No Results" view should appear.

11. Click on a theme under the "Search for a theme instead?" suggestion.

12. The system should now correctly apply only the theme filter and return results
Screenshot 2025-04-22 at 4 54 44 PM
.

13. This fixes a bug where the original query and theme filter were combined, resulting in another "No Results" view.

### DIFF
--- a/src/components/search/detailPanel/iconTag.tsx
+++ b/src/components/search/detailPanel/iconTag.tsx
@@ -41,6 +41,7 @@ const IconTag = (props: Props): JSX.Element => {
       newSubjects = [...currentSubjects, sub];
     }
     if (props.themeOnly) {
+      dispatch(setQuery("*"));
       dispatch(batchResetFilters({
         schema: schema,
         query: "*",
@@ -50,7 +51,6 @@ const IconTag = (props: Props): JSX.Element => {
     } else {
       dispatch(setSubject(newSubjects));
     }
-    
     plausible(EventType.ChangedThemeFilter, {
       props: {
         themes: newSubjects.join(", "),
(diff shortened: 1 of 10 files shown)


mradamcox (approved):
Looks good to me, these were pretty related issues so good to address them all at once. The one thing I want to flag for the future is that some of this inital results + related results combination logic will get complicated once we introduce pagination. Currently, it's no problem to iterate multiple sets of results and filter to only show unique items, but I have a feeling pagination will complicate that process. Anyway, all good for now, thank you for this!

PR #550, opened 2025-04-22
#540 feat: decorative images in markdown

### LINKED ISSUE #540: [STYLE] Facilitate setting some images as "decorative" in Research Guides and reduce width on these images.
Currently, all images in the research Guide body are styled to 100% width:

Image

We need to add a class or other mechanism that will allow authors to indicate that certain images should be have a width of 50% for larger screens, and 100% for tablets and smaller (reusing breakpoints we already have in the code base).

Image

This could be done with a class that is manually added in the CMS body, for example. It should be easy for authors to do.

bodom0015:
## Problem
Currently, images use 100% of the page width

But some of these images are purely decorative and don't need to draw that much of the user's focus

Fixes #540 

## Approach
* feat: allow users to specify which images are decorative - these will be given 50% width (on large screens only)
    * smaller screens should still use 100% width
    * we could also add a "Decorative Image" widget that would add this div/class around the given image, which might ease user experience a bit if folks aren't as familiar with HTML

For example:
```
<div className="decorative-image">
![](/images/prep.svg)
</div>
```

On larger screens, this will render as:
Screenshot 2025-04-22 at 10 52 29 AM

On smaller screens, this will render as:
Screenshot 2025-04-22 at 10 52 12 AM

## How to Test
~~I'm not sure how to offer a test case via Netlify, as I believe DecapCMS will create a new branch/PR for each change
But to use the new class, it would need to use this branch.~~

### Viewing the Preview
1. Navigate to https://deploy-preview-550--cheerful-treacle-913a24.netlify.app/guides/public-transit-equity
2. Scroll down to the section titled "Spatial Approaches to Calculating Public Transit Access at the Community Level"
    * On desktop (> 769px): you should see that the image slightly below this header takes up only 50% of the screen
    * On mobile (<= 769px): you should see that the image slightly below this header still takes up 100% of the screen

### Testing Locally
1. Checkout and run this branch locally: `git fetch --all && git checkout feature/lambert8/support-decorative-images-in-markdown`
4. Navigate to http://localhost:3000/guides/public-transit-equity
    * You should see that images take 100% of the page width here
5. Edit one or more images to be inside of a `<div className="decorative-image">    </div>` block
    * To avoid needing to publish, create a PR, etc, you can simply edit your local `.mdx` file directly
6. Refresh http://localhost:3000/guides/public-transit-equity
    * You should see that images take 50% of the page width here on larger screens
    * You should see that images still take 100% of the page width here on smaller screens

### DIFF
--- a/content/guides/public-transit-equity.mdx
+++ b/content/guides/public-transit-equity.mdx
@@ -79,7 +79,9 @@ We can use certain signs to check if the system treats everyone fairly instead o
 
 ### Preparation of Origin and Destination Points
 
-![](/images/prep.svg)
+<div class="decorative-image">
+    ![](/images/prep.svg)
+</div>
 
 The origin and destinations provide a sense of the population, geographical and multiple resources or opportunities of interest. The spatial location of the origin/destination can be a point location such as bus stops or train stations or the centroid of the geographical boundaries such as a city, census tract, census block group etc.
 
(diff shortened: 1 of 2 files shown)


mradamcox (approved):
@bodom0015 This looks perfect, I'm going to add one commit to put this wrapper around the images in the transit guide, so we can easily see them in the deploy preview here: https://deploy-preview-550--cheerful-treacle-913a24.netlify.app/guides/public-transit-equity

PR #552, opened 2025-04-24
#541 feat: add generic Panel component for markdown editor

### LINKED ISSUE #541: [FEATURE] Add link button/component to Resource Guides to go to Data Discovery app with preset filter
The idea is that there should be a link in each SDOH guide that would go to the Data Discovery application, and show a preset list of results related to that topic. For example, the link in the "Public Transit Equity" guide would lead to: https://sdohplace.org/search?subject=Transportation+and+Infrastructure

@shubhamk008 to design and @bodom0015  to implement. There are different ways to implement, any of them would probably work (likely will depend a bit on the complexity of the design):

- A custom css class that is applied in the preview template and the published post. This would require the author to write some HTML in the body text, but won't be too many guides overall so this could be fine.
- A custom element within the template that gets populated with text content and a search url which are added as fields to the Guide collection in the CMS
- A new component that will appear in the WYSIWYG editor, similar to how we implemented the Toggle Section component. This would need to be hidden for other body fields that use the WYSIWYG editor (Posts and Showcase, etc).

See Design:
https://www.figma.com/design/vD6pC3DsFtGG2aCv4vWXtV/HEROP?node-id=3647-847&t=jbdePqSBbIyHGdNd-4
Image

bodom0015:
## Problem
We need a generic way to link from SDOH guides to the Discovery App

It would be nice if this solution was generic enough to be reused for other things as well

Fixes #541 

## Approach
* feat: added a custom Panel component to the Markdown editor with the following properties:
    * `title`: Header to display at the top of the panel
    * `content`: Markdown body to display on the right side of the panel
    * `link`: URL for the link at the bottom of the panel
    * `linkText`: Text to display for the link at the bottom of the panel
    * `image`: Image to display on the left side of the panel
    * `backgroundColor`: Background color to use for this panel

### DecapCMS Preview
CMS___SDOH___Place_Website

### Rendered Markdown in SDOH&Place
Screenshot_2025-04-24_at_4_42_04_PM

## How to Test
TBD

### DIFF
--- a/content/guides/public-transit-equity.mdx
+++ b/content/guides/public-transit-equity.mdx
@@ -79,7 +79,9 @@ We can use certain signs to check if the system treats everyone fairly instead o
 
 ### Preparation of Origin and Destination Points
 
-![](/images/prep.svg)
+<div class="decorative-image">
+    ![](/images/prep.svg)
+</div>
 
 The origin and destinations provide a sense of the population, geographical and multiple resources or opportunities of interest. The spatial location of the origin/destination can be a point location such as bus stops or train stations or the centroid of the geographical boundaries such as a city, census tract, census block group etc.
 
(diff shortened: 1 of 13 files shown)


mradamcox (approved):
I think this looks fantastic. Because it's a little tough to test these kinds of changes out, I'm going to merge this into main now, and we'll continue to refine it after we can incorporate it into a couple of existing CMS posts.

Issue #499, opened 2025-03-12
[FEATURE] Add legend/indicator about map scale when a boundary show up in map


pengyin-shan:
In the map view, when the boundaries show up, I think it could be beneficial to have a legend indicating what scale that is.

mradamcox:
Scale bar added.

PR #563, opened 2025-05-13
feat: add scale control #499

mradamcox:
Quick addition of a basic scale control (using the bindings through react-map-gl: https://visgl.github.io/react-map-gl/docs/api-reference/maplibre/scale-control.


pengyin-shan (approved):
See a scale at the bottom left corner of the map. Approve this PR

Issue #567, opened 2025-05-13
[BUG] Check all filters to make sure they runs correctly with query


pengyin-shan:
After navigation, the "related results" are not properly fetched when the user applies filters. Integrate the filters into the fetching logic. Revisit this later to discuss potential improvements to the recommendation mechanism.

PR #569, opened 2025-05-13
#567: Apply filters to the related result

pengyin-shan:
This should apply filter actions to the Keyword Search. Unfortunately, for AI Search, it appears to require additional work. We can revisit this later, as the search side of AI Search with filters triggers a re-fetch for the LLM, potentially resulting in different outcomes.

## How to Test
Run a search with filters. Try different interaction sequences—applying filters first, searching first, or any other combinations—to ensure it functions as expected.

### DIFF
--- a/src/middleware/actionConfig.ts
+++ b/src/middleware/actionConfig.ts
@@ -102,7 +102,7 @@ export const actionConfig: Record<string, ActionConfig> = {
   "search/setSort": {
     param: "sort",
     syncWithUrl: false,
-    requiresFetch: false,
+    requiresFetch: true,
     isFilter: false,
   },
 
--- a/src/services/SearchService.ts
+++ b/src/services/SearchService.ts
@@ -147,6 +147,7 @@ export class SearchService {
   async fetchRelatedResults(
     validSuggestions: string[],
     usedQuery: string,
+    filterQueries: Array<any> = [],
     skipCache: boolean = false
   ): Promise<any[]> {
     const uniqueSuggestions = validSuggestions.filter(s => s !== usedQuery);
@@ -155,9 +156,8 @@ export class SearchService {
       if (suggestionManager.hasSuggestion(suggestion)) continue;
       try {
         suggestionManager.addSuggestion(suggestion);
-        const { results: suggestionResults } = await this.queryBuilder
-          .generalQuery(suggestion)
-          .fetchResult(undefined, skipCache);
+        this.queryBuilder.combineQueries(suggestion, filterQueries);
+        const { results: suggestionResults } = await this.queryBuilder.fetchResult(undefined, skipCache);
         suggestionManager.removeSuggestion(suggestion);
         
         if (suggestionResults && suggestionResults.length > 0) {
--- a/src/store/slices/searchSlice.ts
+++ b/src/store/slices/searchSlice.ts
@@ -184,9 +184,8 @@ export const fetchSearchAndRelatedResults = createAsyncThunk(
             suggestionManager.addSuggestion(suggestion);
             const queryBuilder = new SolrQueryBuilder();
             queryBuilder.setSchema(schema);
-            const { results: suggestionResults } = await queryBuilder
-              .generalQuery(suggestion)
-              .fetchResult(undefined, false);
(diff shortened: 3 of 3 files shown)


mradamcox (changes requested):
It seems like there is still some bleed over from the suggested results into the main query (or something like that). Here's what I'm finding:

1. select "block group" filter (14 results)
    image
2. search for keyword "poverty" (20 results)
    image


I would expect a search for a keyword to be treated as an AND clause alongside the filters, so there would be fewer results. The results after step 2 include items that do not have Block Group spatial resolution.

One stranger thing is that clearing the keyword search (clicking the x in the box) produces a results list with 31 items:

image

I would expect this to simply rerun the existing Block Group filter query and return 14 items. Instead, it seems like suggested terms are processed after the field is cleared and they contribute to more results.

If I refresh the page (which retains the Block Group query through URL params) then the expected number of items are returned.

Ultimately, I'm not sure if it's worth working more on this particular PR if a deeper overhaul of the search query strategy is needed.

pengyin-shan:
> It seems like there is still some bleed over from the suggested results into the main query (or something like that). Here's what I'm finding:
> 
> 1. select "block group" filter (14 results)
>    image
> 2. search for keyword "poverty" (20 results)
>    image
> 
> I would expect a search for a keyword to be treated as an AND clause alongside the filters, so there would be fewer results. The results after step 2 include items that do not have Block Group spatial resolution.
> 
> One stranger thing is that clearing the keyword search (clicking the x in the box) produces a results list with 31 items:
> 
> image
> 
> I would expect this to simply rerun the existing Block Group filter query and return 14 items. Instead, it seems like suggested terms are processed after the field is cleared and they contribute to more results.
> 
> If I refresh the page (which retains the Block Group query through URL params) then the expected number of items are returned.
> 
> Ultimately, I'm not sure if it's worth working more on this particular PR if a deeper overhaul of the search query strategy is needed.

Ah I see, thanks @mradamcox! so if the filter is applied after the search, the filtered results are currently correctly reflected in the "related results." However, doing it the opposite way, as you showed, fails to trigger that behavior. I think it might be related to additional filter-triggering consequences that I need to dig into. And yes, let's hold off on merging this PR for now.

mradamcox:
This brings up another thought: Maybe as we improve the AI search mode, we have less need for this pattern of adding suggested results at all? I feel like a "can't find what you need, try the AI assisted search to broaden your results" message would be entirely reasonable if no (or very few) results come from the keyword search.

pengyin-shan:
> This brings up another thought: Maybe as we improve the AI search mode, we have less need for this pattern of adding suggested results at all? I feel like a "can't find what you need, try the AI assisted search to broaden your results" message would be entirely reasonable if no (or very few) results come from the keyword search.

Exactly, I was thinking the same. Let's review this mechanism sometime and discuss if we should turn to AI's help on this. Currently, it seems to add an extra burden that is making the search more complex

mradamcox:
> so if the filter is applied after the search, the filtered results are currently correctly reflected in the "related results." 

Ah, well I didn't try that, but yes that appears to be the case:

1. search "poverty" alone
image

2. now add spatial resolution filter
image

These 3 look like properly matched results.

mradamcox:
> Currently, it seems to add an extra burden that is making the search more complex

Yeah, I hadn't looked in the network tab for a while, but triggering all of those extra requests to Solr is pretty cumbersome. Maybe a _single_ extra query could be sent that is an OR of the top 5 suggested items, or something like that.

pengyin-shan:
> > Currently, it seems to add an extra burden that is making the search more complex
> 
> Yeah, I hadn't looked in the network tab for a while, but triggering all of those extra requests to Solr is pretty cumbersome. Maybe a _single_ extra query could be sent that is an OR of the top 5 suggested items, or something like that.

Yes, the original recommendation methods significantly increase search time due to multiple fetches, compared to a single fetch. I'm considering configuring Solr to return more "related" results in a single fetch, so we can eliminate the need for extra queries in the future. This would also allow us to handle server-side pagination and sorting 🙌

pengyin-shan:
close this as we are moving to a non-related-result solution

Issue #571, opened 2025-05-15
[TASK] Remove the old recommendation mechanism since AI has been adopted


pengyin-shan:
Now that we have implemented AI-Inspired search, the legacy recommendation mechanism (triggered by hitting the 'suggest' endpoint for every searched term, regardless of whether it originates from a keyword search or is recommended by the LLM) is adding unnecessary load to the app. This mechanism performs multiple fetches and requires additional filtering and sorting, which not only increases page load times but also introduces a higher risk of errors.

With the transition to AI-Inspired search, it's time to streamline the search process by removing this legacy mechanism (see discussions in [PR #569](https://github.com/healthyregions/SDOHPlace/pull/569)). This change will allow us to perform a single fetch to retrieve all results, simplifying server-side sorting, filtering, and pagination.

In addition, I will update the help text for Keyword Search mode to indicate that it expects exact matches and we are actively working on ontology enhancements to improve relevance in the near future. For AI-Inspired Search, I will adjust the prompt to return terms that better align with our current dataset.

After this update, please revisit issues #556, #567, and #555 to ensure filter and sort operations function correctly. Finally, be sure to adjust the upcoming routing PR to reflect these changes as well.

PR #572, opened 2025-05-15
#571, #567: Remove related result fetching and fix filter

pengyin-shan:
This PR addresses issues #571 and #567. I’ve removed the “related results” feature from both keyword search and AI search. The corresponding methods are commented out (with notes), so they can be easily re-enabled or referenced in the future if needed. I also fixed the filter behavior to reflect the selected filters more accurately.

As a result of removing the recommendation feature, users may encounter the “No Results” page more frequently. This is expected, and we can later analyze search terms that trigger these empty results using Plausible to improve our ontology and Solr configuration.

Additionally, I simplified the highlight message and updated the AI search to return a fixed number of recommended terms, which can help generate more relevant results.

## How to Test
1. Run the app and perform a search. Then apply filters — the result count and displayed results should match the selected filters. Test all filters to ensure they work correctly.

2. Try the reverse flow: apply filters first, then perform a search. The results should be consistent with expectations.

3. Explore the app to verify that all previous functionality remains intact. Note: I’ve added a message to the AI search mode to inform users that filters currently do not apply in that mode (I can block filters in AI search mode entirely if that provides a better experience). Future work will include persisting AI-recommended terms to prevent loss on refresh and enabling server-side filtering based on those terms. This will be addressed in a future build and is not part of this PR.
Screenshot 2025-05-15 at 6 34 51 PM


mradamcox (approved):
This is great, I think it makes a lot of sense to simplify the related search suggestion logic and the speed up basic search results, now that the AI search can effectively mimic (and do a better job of) the related search experience.

Issue #161, opened 2024-05-29
SEO Audit - Prioritize Top-level pages


mradamcox:
Our priority for SEO is to make sure that all of our top-level pages are well-indexed, so the first step should be an audit of what NextJS gives us out of the box. Once an audit has been carried out, results should be written here, and, if necessary, more specific task tickets can be made.

Topics to consider:

- Are all meta head tags properly populated?
- Do we need a sitemap.xml file?
- Do we need something like Prerender.io? Other options for static page generation?

There are more topics to consider, I just don't know what they are... that's what the audit is for!

mradamcox:
While working on #627, I've learned a lot more about how we should have things set up. There are a few different protocols that would be good to support within the site, and our initial implementation of the Decap CMS-based "news" blog included many of these, but it was sequestered in that part of the app and hadn't been distributed across all pages like it should have been (this is ultimately the cause of #627).

### Types of tags and where we can use them

To summarize what seem to be the most important meta tags and attributes to have:

#### Viewport tag

Seems like we can use this default value on all pages:

```
<meta name="viewport" content="initial-scale=1, width=device-width" />
```

#### Basic meta tags

These need to be on every page, content is dependent on the page itself.

```
<title>{title for this page}</title>
<meta name="description" content="{description of this page or the site overall}" />
<meta name="keywords" content="keyword1,keyword2,keyword3" />
<meta name="author" content="{name of author of page}" />
<link rel="canonical" href="{current url to this page}" />
```

#### OpenGraph tags

- ref: https://ogp.me/
- ref: https://developers.facebook.com/docs/sharing/webmasters/#markup

These are used by Facebook, LinkedIn, and perhaps others. It is a more robust set of tags, and it varies more depending on what page is rendered. There is a `og:type` property with a few valid options, and these determine what the other tags should hold. As far as I can tell, the only relevant types for us are:

- `website`: This is the default type according to Facebook docs, and we don't need to explicitly set it. Use this for all standard pages, like home page, about, etc. A full set of these tags is pretty basic:
    ```
    <meta property="og:title" content="{same as meta title above}"  />
    <meta property="og:url" content="{same as canonical url above}" />
    <meta property="og:description" content="{same as meta description above}" />
    <meta property="og:site_name" content="{overall name of website}" />
    <meta property="og:image" content="{full url to image for unfurl (provide default logo as well}" />
    ``` 
[...]

PR #632, opened 2025-07-08
consistent OpenGraph metadata across static site content

### LINKED ISSUE #627: Fix all meta tags to be uniform across all pages
## problem

Certain unfurling behaviors, like on Linkedin, rely on the opengraph or og:* tags like

```
<meta property="og:url" content="your website url" />
<meta property="og:title" content="the title that shows up in the unfurled preview" />
<meta property="og:description" content="the description in the preview" />
<meta property="og:image" />
```

to determine what shows up in the small "unfurl" cards that are generated when someone puts a link in a post. However, we don't have those tags consistently set across all pages in the site.

This is related to #161 

## solution

We need to refactor the way the main Head element is created and shared across all pages to ensure that every page has the proper set of meta tags. Currently, there is a basic set of tags on most pages, but only those that inherit architecture from the original set up of the blog have a full complement of tags, including `og:*` and oEmbed.

Plus, it would be good to do a little more research to see if there are other protocols we should include as well.

bodom0015:
## Problem
The default OpenGraph metadata appears to be strangely selective about which text appears in a preview on some sites, such as LinkedIn

Fixes #627 

Screenshot 2025-07-08 at 4 35 31 PM

## Approach
* wip: initial testing using LinkedIn inspector

TODOs
* figure out why the current text is being selected - maybe `<h3>` element?
* replace this with proper OpenGraph meta tags on each page
    * reuse/refactor existing components where possible

## How to Test
1. Navigate to https://www.linkedin.com/post-inspector/inspect/https:%2F%2Fdeploy-preview-632--cheerful-treacle-913a24.netlify.app
    * You should see a preview of how this site will be rendered when linking to it


mradamcox:
I got a little carried away with researching these different tags, and ended up with a bit of a refactor that is described in more detail here: https://github.com/healthyregions/SDOHPlace/issues/161#issuecomment-3104152381

Just to reiterate: https://github.com/healthyregions/SDOHPlace/issues/627 was not so much about about a single header (the `h3` pretty far down the home page was being picked up by LinkedIn as a title) but more about the fact that there were no OpenGraph elements at all on the home page (or other standard sub-pages). I assume that `h3` was being picked up because in the absence of good meta tags the first `<h*>` element was selected... the fact that we don't have any other h elements on the home pages is itself a different problem... At any rate, I ultimately decided to take this opportunity to put a lot more time into #161 and better understand how all of these tags can be used for SEO.

mradamcox:
Update: Decided to make the `BasicPageMeta` and `ArticleMeta` components mutually exclusive. So they both contain the same base content, but the Article component has extra stuff just for articles. I think this is a bit cleaner and easier to manage.

PR #638, opened 2025-07-17
feat: dedicated page for Research Outputs

### LINKED ISSUE #617: [FEATURE] Dedicated page for Research outputs
We would like to explore creating list of Research / Reports output from the SDOH&Place project
This is the hidden link that was missing from #580 (commented out)

We would like to source this page from DecapCMS if we can create a schema for Research entries, just like we do for News/Showcase/Guides, and can borrow a lot of the style / structure from the existing "SDOH Guides" page (see #539)

Something like the following might work as a basic schema:
* title - title to display in the list (e.g. A Guide to Human-Centered Design)
* url? - can we assume that all entries have an external URL associated with them? always PDFs?
* doi? - can we assume that all entries are published works?
* date? - should this be date it's added to CMS? separate publish date?
* body? - content to display when a single entry is chosen. provide a link to the full resource at the end
* ... are there any other properties we are interested in?

Open Questions:
* do we want a page that gives a short description/preview of each report before linking to the full text?
    * alternatively, clicking a title in the list could simply open a new tab with a PDF viewer
        * I would prefer to use the built-in browser PDF viewer in this case
* should this replace the "Featured" section? Is the Featured section always going to feature the most recent Research/Report item?
    * perhaps it might be desirable to have both a "Featured" section for general features, as well as a section that also highlights the most recent research item?
    * or maybe the "Featured" section can simply be updated when we release a new Research item to keep it flexible and allow us to use it to announce other things (e.g. Symposium, call for participation, etc)?
    * see #580 
    * see #581
    * this part might be a separate ticket after the initial work

bodom0015:
## Problem
We do not currently have a CMS-driven page where our Research Outputs can live

For example: `A Guide to Human-Centered Design`

Fixes #617 

## Approach
* feat: create basic schema for research-outputs
    * based on the schema defined in #617
* feat: added frontend lib code to fetch research entries
* feat: added new route/basic layout for `/research` which lists all Research Outputs
* ~~feat: added new route/basic layout for `/research/[slug]` which show details / links for a particular research item~~

Screenshot 2025-07-18 at 2 04 48 PM

Screenshot 2025-08-07 at 3 20 36 PM

### TODOs

- [x] frontend lib code to fetch research entries
- [x] basic page layout + routing based on slug:
    - [x] list all research-outputs present in the system
- [x] when clicking a single item in the list, popup a dialog box that:
    - [x] shows Description/body, render as markdown
    - [x] offers a link to the related items
    - [ ] ~~support multi-media~~ we need a way to specify unique ink label per-file resource, and this does not seem to be possible when using `type: file`

## How to Test
1. Navigate to https://deploy-preview-638--cheerful-treacle-913a24.netlify.app/research
    * You should see a list of all (1) Research Outputs


mradamcox:
> Can we add some space between the timestamp and the description? I'm unable to make any space changes in the code upon inspection, so start with a 1rem gap and then let me know.

image

mradamcox:
I really don't understand it, but seems like the deploy preview subdomain is the problem, from Netlify I can get straight to a different link for each specific build. For example, here is the latest one: https://688923cb731e490bb442ef44--cheerful-treacle-913a24.netlify.app/

mradamcox (changes requested):
Overall this is looking really good, here are a couple of the final updates we have discussed just to keep track:

- [x] Add `1rem` margin as mentioned above
- [x] Finalize link handling for resources
- [x] Update the intro text
    > The SDOH & Place team regularly releases reports, research articles, and related products. Learn more about our work to advance the measurement and understanding of the Social Determinants of Health and structural drivers of health!
- [ ] Add an Author field, type: string / free text.
- [ ] Remove subpage entirely
- [ ] Remove click / link from full item in list
- [ ] Add link to resource, see home page for example, directly below item text description

Issue #641, opened 2025-07-18
AI Search is not working on the production server due to Netlify Edge Function


pengyin-shan:
Resolve this issue and add an error-handling mechanism to provide clearer feedback to the user.

PR #643, opened 2025-07-18
#641: Resolve the Netlify Edge function failure over the AI-Search in Production Server Only

pengyin-shan:
This PR addresses issue #641. The primary purpose of this PR is to demonstrate our fallback strategy when AI-powered search fails for any reason. To simulate such a failure, I intentionally renamed the chat-search endpoint to cha-search. Once this PR is approved, I will revert it back to chat-search before merging.

We do need to deploy this to production to verify whether the fix works, as AI search currently functions correctly on both the local and dev environments. The fix is a simple update to `netlify.toml`, and I suspect the previous configuration may have worked for local/dev paths but is no longer compatible with production.

@mradamcox, let me know your thoughts on the best deployment strategy. One approach could be to first create a minimal PR that only updates `netlify.toml`, merge it into main, and deploy. If that resolves the issue, we can then merge and deploy this PR as a fallback strategy for any future AI failures.

## How to Test
1. Assume that AI search fails (e.g., due to an unreachable OpenAI API or a Netlify Edge Function error).

2. Go to the search page and switch to AI mode. Enter: “What impact does housing stability have on the health outcomes of low-income populations?”

3. You should still see results coming in. The “Inspired by your search” message should indicate that AI search failed and we’re using local semantic parsing as a fallback.
Screenshot 2025-07-18 at 5 12 46 PM

4. The fallback semantic parsing uses a simple dictionary array defined in `sdohSemanticDatabase.ts`. This is an early version and can be expanded later. Its main goal for now is to ensure we have a basic backup in place.

5. To test locally, you can change `await fetch(${baseUrl}/cha-search` back to `await fetch(${baseUrl}/chat-search` in `SearchUtils.ts`. AI search should then work locally. However, since the problem only appears on the production server, we need to deploy the updated `netlify.toml` file to production in order to confirm whether the issue is resolved.

@mradamcox @Makosak  Let me know your thoughts!


mradamcox (approved):
@pengyin-shan I think we can merge this through (after we update the `cha-search` path of course :) ). I've tested locally and this definitely provides a good backup strategy in case there is an error with the netlify function.

Issue #620, opened 2025-06-27
"Clear ALL" and other filter behavior linking to map


pengyin-shan:
Report: "Clear All" will require to clear sort and filter along with any map filter applied but the search bar remains with the search keyword. Moreover in the new design individual filter/sort chips will be seen.

This is a known issue related to connecting the map API with the front end. We've tried several strategies to trigger the "Clear map input box" action via the API—through the "Clear ALL" button instead of relying on the user's manual input—but have seen limited success. We may need to further explore the map API or consider replacing it with a different one.

pengyin-shan:
May need support from Map API and re-design if the map api cannot achieve the 'Clear All' feature

mradamcox:
Yeah, we definitely need to try this once more with some of the methods on the geocoding control, but I do recall us trying this out once already with no luck. Really seems like something that should be doable...

shubhamk008:
I'm also planning on renaming "Clear all" to "Reset search", just so it is abundantly clear that it will reset all search fields/filters being applied. But if we are unable to make this master reset work, then let me know.

pengyin-shan:
> This should be resolved through the linked PR. For what it's worth, I completely agree with "Reset search" over "Clear all". [@pengyin-shan](https://github.com/pengyin-shan) and [@shubhamk008](https://github.com/shubhamk008) Should I go ahead and make that change on this current PR, or just wait until we get it merged and do it after?

Let's go ahead and merge the PR to get the "Clear All" functionality working first and address the confusion some users had in the Time I survey. After that, I suggest we monitor the results of the Time II survey to see if this change helps users navigate more easily or improves the experience in other ways. If no one raises concerns after that, we can then consider moving to a "Reset Search" approach and evaluate user feedback in the Time III survey.

Issue #625, opened 2025-06-27
Map API-Related Task


pengyin-shan:
_Request 1: # Some well-known places do not appear in the map search options. (e.g., Buffalo, NY). I also noticed that the highlighting does not happen for zip codes._

This may need to be configured on the API side to see how the city options are defined, since this part is not controlled by us. 

For the highlight feature, once I confirm the field used for ZIP codes, they can be highlighted as well. I may also need to adjust the highlighting options to support numeric-only values

pengyin-shan:
Request: (map) Speed can be faster.

pengyin-shan:
@mradamcox these requests from the survey related to the map. These look like more of feature requests, except "Some well-known places do not appear in the map search options. (e.g., Buffalo, NY)". I assigned you these map tasks since you are more familiar with them, but feel free to update anything as you need.

mradamcox:
@pengyin-shan Sounds good. Some of these things may be able to be addressed, but there are some limitations based on the underlying service we use for the map search, so some of this just can't be fixed at this time unfortunately. Will give a more detailed report when I'm able to work on this.

mradamcox:
Just a bit more context about the zip codes, I did encounter this limitation [before when I was styling the map](https://github.com/maptiler/maptiler-geocoding-control/issues/77#issuecomment-2549459309.) and unfortunately there isn't a lot we can do about it easily. The geocoding api we use does not return polygon geometries for zip codes like it does for states or counties or municipalities--it returns points or multi-points (which actually makes a lot of sense, because these are zip codes, not zip code tabulation areas...).

The Buffalo, NY issue is really interesting though and warrants more investigation.

mradamcox:
Looking further into the Buffalo, NY example, the problem was that I restricted the geocoding control to these types of places (see [reference](https://docs.maptiler.com/cloud/api/geocoding/#PlaceType)):

```
types: ["region", "county", "postal_code", "municipality"]
```

However, Buffalo, NY is none of those place types (turns out it is listed as a `joint_municipality`) so I expanded the types list significantly.

```
types: ["region", "county", "postal_code", "municipality", "municipal_district", "joint_municipality", "joint_submunicipality", "locality", "neighbourhood"]
```

The country is still limited to `us` though.

PR #647, opened 2025-07-27
620 clear map with other filters

mradamcox:
This PR implements a bit of a refactor and cleanup to the map code, along with closing #620. I got some [help with the geocoding control](https://github.com/maptiler/maptiler-geocoding-control/issues/96#issuecomment-3124721603) to achieve the clearing function, which comes down to this snippet:

```
gc.current.setOptions({apiKey:apiKey, clearOnBlur:true});
setTimeout(() => {
    gc.current.clearMap()
    gc.current.focus() 
    gc.current.blur() 
    gc.current.setOptions({apiKey:apiKey, clearOnBlur:false});
}, 1000);
```

However, I was having a lot of trouble figuring out how to expose the control instance in a way that these methods could actually be called in a way that affected the interface. What I ended up deciding to do was remove one layer of abstraction that we had been using, the `react-map-gl` library, and rely instead directly on `maplibre-gl`. (We were already using MapLibre, but accessing through the `react-map-gl` wrappers, and I think those React component wrappers were making it too hard to get to the level of control that we really needed--at least with my knowledge of React.)

So, in the end we now have one less dependency (yay!) and are only dealing with MapLibre classes directly, not mixing them with the React components that the other library provides on top of them. There are a few parts of the map code that are slightly more verbose now, but it's a net reduction in code, and overall I think it actually makes an easier-to-understand component.

After that was all handled, hooking up the control clearing to the "clear filters" button was straightforward, using a new property of the search.ui slice.


mradamcox:
> Do you think we should also reset the zoom level back to the full US view? Not necessary, just something to consider.

Yeah, that's a good idea, that will be an easy addition to make. I'll reset the map to the initial bounds that are used to create the map on load.

mradamcox:
I've also added a fix for part of #625, the problem with Buffalo, NY not showing up in the search control.

The deploy preview build failed the first time, but I think it was a netlify problem because I re-triggered it there and it has completed. Still seeing if I can re-trigger the checks that appear here on the pull request...

pengyin-shan:
> I've also added a fix for part of #625, the problem with Buffalo, NY not showing up in the search control.
> 
> The deploy preview build failed the first time, but I think it was a netlify problem because I re-triggered it there and it has completed. Still seeing if I can re-trigger the checks that appear here on the pull request...

Thank you @mradamcox ! Just that when testing this branch, I saw this after clicking the 'Clear All' button:
Screenshot 2025-07-29 at 10 26 10 AM

mradamcox:
The problem was this dependency array:

```
useEffect(() => {
        if (gcRef.current && !geosearchSelection) {
            gcRef.current.clearMap()
            mapRef.current.fitBounds(props.initialBounds)
            gcRef.current.setOptions({apiKey:apiKey, clearOnBlur:true});
            setTimeout(() => {
                gcRef.current.focus()
                gcRef.current.blur()
                gcRef.current.setOptions({apiKey:apiKey, clearOnBlur:false});
            }, 1000);
        }
   }, [geosearchSelection])
```

To please the linter I had added `props.initialBounds` to the dependency array just before commiting, but this causes the effect to loop and get to max recursion depth. For now, I've just removed it from the array and will live with the warning. May be something to look into later.

pengyin-shan (approved):
Tested both the "Clear All" feature and the location dropdown — everything works great. This PR will also help simplify future map operations. Approving the PR.

PR #648, opened 2025-07-27
docs: add note about community asset layer accuracy

### LINKED ISSUE #626: Data verification/update task
_Request: The Parks Community Asset seems like it has some errors. It includes things like nurseries and businesses. I also noticed some parks were in the wrong spot and some long-established city parks were omitted_

mradamcox:
Added a note to address one part of #626:

image

@shubhamk008 Let me know how this should be changed to make it ready to go.


shubhamk008:
Looks good. Once I begin working on personification then I'll revisit the UX copy and suggest changes if any.

Issue #649, opened 2025-07-29
Create a env.example file to run AI search locally


pengyin-shan:
Since we previously switched between uiuc.chat and OpenAI, I’ll create an env.example file to document the required environment variables for the AI search setup.

PR #650, opened 2025-07-29
#649: Update env example file and move the variable explanation from README to env file

pengyin-shan:
This PR addresses #649. As discussed in Slack, I’ve also moved the environment variable explanations into this file to make the setup easier.

Users should now be able to follow the instructions and run the Discovery app with AI search locally. Note that I’ve kept the `uiuc.chat` variables marked as required for better error handling in the app. However, if the OpenAI API is used instead, these can simply be set to dummy values (e.g., xxx), which I’ve noted in the comments.


mradamcox (changes requested):
Looks great, I think it's a good idea to have the notes in the example file itself. Are the `SOLR_USERNAME` and `SOLR_PASSWORD` vars actually used though? I can't see where they are referenced in the code. If not, let's remove them from here and then I'd say we're good to merge this in.

mradamcox (approved):
Looks good, thanks!

PR #642, opened 2025-07-18
Update Guides “public-transit-equity”

bodom0015:
Automatically generated by Decap CMS

Starts to address #637

## How to Test
1. Navigate to the CMS preview: https://deploy-preview-642--cheerful-treacle-913a24.netlify.app/guides/public-transit-equity
2. Scroll all the way down to the "About the Author" section
    * You should see the new panel-based generic approach here
    * Please keep note any potential styling improvements, as this may be the first time that we're actually utilizing this component - see below for a list of subtle things that I've already noticed, and feel free to comment more

### Styling

Manual HTML (Old):
Screenshot 2025-07-18 at 2 38 16 PM

Panel Component (New):
Screenshot 2025-07-18 at 2 37 02 PM

Notable styling differences (@shubhamk008 may have opinions here as well):
- header font size needs a slight increase
- header needs slight top margin
- image should be rounded
- image could be slightly larger
- image needs an orange border (`border-salmonpink` => `rgb(255 156 119/var(--tw-border-opacity))`)
- link + link text should be optional (hide `<a>` if no link or text specified)
- shrink bottom padding when no link is present
- increase left padding on body text

for each of the above, the question is always: "will all panels need this? or is this an option for a specific case?"

e.g. maybe `roundedImage` or `imageBorderColor` becomes a property on the Panel component? I think most of the other items here make sense for all panels, but these are ones where I'm not sure. Or maybe I'm overthinking the styles right now, and we can deal with this part later 🤷‍♀️ 

also noting that the more options we add, the more complex the  RegEx to parse it becomes and it's already pretty ridiculous 😅 for example, this is why we simply assume that a link will always be present, because trying to handle both cases (e.g. having a link and having no link) in a single RegEx is not trivial

Issue #637, opened 2025-07-15
Add "About the Author" to new research guides - use Panel block in CMS


mradamcox:
Two of the research guides already have html blocks for "about the author", built with this code block:

```
<div style={{padding: "1em 1.5em", display:"flex", flexDirection:"column", borderRadius:"1em", backgroundColor:"#ffe5c4"}}>
  <h4 style={{marginTop:0, marginBottom:".5em"}}>About the Author</h4>
  <div style={{display:"flex"}}>
    <div style={{minWidth:"160px"}}>
      <img className="rounded-full border-4 border-solid border-salmonpink" src="/images/headshot-mg.png" />
    </div>
    <p style={{marginLeft:"2em", marginTop:"0"}}><strong>Mandela Gadri</strong> is a Ph.D. student at <strong>University of Illinois at Urbana-Champaign (UIUC)</strong> with an MS in community and regional planning from Iowa State University and a BS in urban planning from Kwame Nkrumah University of Science and Technology, Ghana. He also has a certificate in GIS and worked as a planner with the city of Cedar Rapids, Iowa prior to joining UIUC. Mandela is interested in GIS and transport poverty.</p>
  </div>
</div>
```

Which is rendered like this:

<img width="1984" height="471" alt="Image" src="https://github.com/user-attachments/assets/8b4c42f0-5f85-4b5c-8aaa-81f3d570a99f" />

However, we now have a new block available to the rich text editor, which approximates this same design but has more flexibility. We should implement this Panel block for all About the Author sections from now on, and retrofit existing posts to match that pattern as well.

- [x] Replace the existing About the Author blocks in this guide: https://sdohplace.org/guides/public-transit-equity with the new Panel block.
- [x] Add an About the Author block to each of the other three guides
  - [x] https://sdohplace.org/guides/greenspace-access
  - [x] https://sdohplace.org/guides/area-deprivation-indexes
  - [x] https://sdohplace.org/guides/measuring-food-access-on-a-neighborhood-level 

Use this hex for the background color of the block: `#ffe5c4`

bodom0015:
@camring2 has provided the bios for 2 of our guide authors 🙌 

**Area Deprivation Indices** (#644):
[...]

PR #654, opened 2025-07-30
style: improve styling for sdoh-panel image + text in CMS content

bodom0015:
## Problem
`sdoh-panel` lacks proper styling, and could use some minor improvements

Related to / discovered during #642 
Partially addresses / related to #637 

Current design specification:
image

Original implementation - manual HTML:
image

Current implementation - CMS-driven Panel:
Screenshot 2025-07-30 at 6 17 58 PM

We would like the implementation to more closely match the design (ignore background color - the panel in the designs is being used for a different context and uses a different color)

## Approach
* wip: discussed using/enhancing this pattern with Shubham
    * the existing styling applied to the "About the Author" images is not general enough to work for all images, and the options to describe the are too broad/complex to capture as a regular expression (e.g. round image, border color, etc)
    * another possibility is to have the border as part of the image, but this would require us to adopt the same border on the Team page or maintain separate images 🤔 
    * compromise: remove the border, always round the image border
* style: implement some simple styling improvements that should apply to both the CMS preview and the rendered version on our site (hooray shared CSS :tada:)
    * slightly larger header font, increased margin
    * image should always be rounded, no border or shadow
    * top-align image with body text
    * `sdoh-content` should have some extra padding to X-offset it from the header slightly
    * remove extra bottom padding

Here is how it looks in the CMS preview (compare these with the initial states above, to provide notes/feedback):
Screenshot 2025-07-30 at 6 15 51 PM

However we can't preview this new styling yet with the About the author content because it is in a different branch (see #642)
* CMS changes are being merged to `publish`
* styling changes are being merged to `main`
* next time we perform a publish step (to sync them), we should see both? I think that's how that works anyways 😅 

## How to Test
1. Navigate to https://deploy-preview-654--cheerful-treacle-913a24.netlify.app/admin/index.html#/collections/guides/entries/public-transit-equity
    * You should see the CMS preview of the Public Transit Equity guide
2. Scroll down to the "About the Author" panel/section
    * You should see the new styling applied as described above

### DIFF
--- a/public/styles/posts.module.css
+++ b/public/styles/posts.module.css
@@ -203,7 +203,7 @@
     margin-top: 0.5rem;
   }
   .sdoh-panel-content {
-    padding: 0 1rem;
+    padding: 0 0.5rem;
     display: flex;
     flex-direction: row;
     margin: 1rem 0 0 0;


mradamcox (approved):
Looks good to me. It's good to get this component updated to handle that section in the CMS.

Issue #623, opened 2025-06-27
Misunderstanding about the map usage


pengyin-shan:
_Request 1: I was unable to get "Show on map" to display any of the scores, only the boundaries (i.e. census tract borders for National Environmental Public Health Tracking Network)_

_Request 2: The map tends to be hyper sensitive. Meaning, it moves with minimal touch/click_

_Request 3: # No specific bugs or errors but filter by location is rather clunky, and a lot of times for different selections, I do not see overlays on the map._

May need design to illustrate the usage of the map features, see comments in Notion https://www.notion.so/Bug-Tracker-from-Survey-21509724d38c801abb3dd834c9d243d9

pengyin-shan:
Also with #622

pengyin-shan:
Request 4: The only problem I have is show on map…When some datasets have "show on map" available, it is not clear to me what information or variable is being mapped/visualized as all polygons seem to have the same color? or maybe I miss
something there?

Request 5: Previously mentioned, the "show on map" links don't appear to be working. I think the links should
switch automatically to the map view

pengyin-shan:
Request 6: Access to Jobs and Workers Via Transit, Crimes by County (NaNDA), Daily County-Level Ozone
Concentrations, Daily County-Level PM2.5 Concentrations, EJScreen, Food Access Research
Atlas, Food Environment Atlas, Global Horizontal Irradiance, Housing + Transportation Index,
National Environmental Public Health Tracking Network, Palmer Drought Severity Index, Park
10-min Walk Service Areas (ParkServe), Social Vulnerability Index, UGS Extractor, Ultraviolet
Irradiance, Uniform Crime Reporting Program, USGS Land Cover: All have an option for "show on map" but nothing is displaying on the built-in map. I also think clicking "show on map" should switch back to the map view.

pengyin-shan:
Request 7: Please make the "show on map" button clearer!

pengyin-shan:
Request 8: Some of the map views can be confusing. For example, under Ultraviolet Irradiance, when I go to
the map view, it only shows blank census units without any data visualization. I tried other datasets,
such as National Environmental Public Health Tracking Network, as it remained the same census
units. I am unsure where I can select variables to be added to the map.

mradamcox:
I agree that "Show on Map" is a confusing label, "Show coverage" or "Show spatial coverage" is a more succinct description of the action that the button performs.

[...]

PR #658, opened 2025-08-05
Button label update

mradamcox:
- As referenced in this comment: https://github.com/healthyregions/SDOHPlace/issues/620#issuecomment-3133625923, this PR updates the "CLEAR ALL" button label to now read "RESET FILTERS".
- As referenced in this comment: https://github.com/healthyregions/SDOHPlace/issues/623#issuecomment-3053102999, update the "Show on map" / "Remove preview" buttons for each item to now read "Show coverage" / "Remove coverage".
- As referenced in that same thread, I've also modified the behavior of the "Show coverage" button so that if the details panel is open, it will be closed when "Show coverage" is clicked.


pengyin-shan (approved):
Tested all three behaviors and they worked well. Approving this PR

PR #696, opened 2025-11-12
Fix site name in unfurl

mradamcox:
Currently, the `&` in the site title is replaced with the HTML entity `&amp;` when Slack unfurls a link to our site.

image

Hopefully this can be fixed, making a pr to test some strategies.


mradamcox:
Ultimately I found this is a bug in Slack, where the _preview_ of the unfurl shows the html entity, but once the message is sent, then the unfurl content does display properly. Reverted to the original content.

Also, updating some wording on the main guides page upon request.

PR #701, opened 2025-11-18
698 remove search page

mradamcox:
This PR removes a lot of content, everything that was related to the Data Discovery application.

- a number of the libraries from package.json that are no longer needed
- all tsx files that contributed only to the /search page
- redux and store framework
- netlify edge functions and related package.json scripts
- all mention or reference to .env, because it is no longer needed
- chat prompts
- metadata schema json files
- the API routes that we weren't really ever using anyway

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,5 @@ services:
     image: herop/sdoh-homepage
     build:
       context: .
-    env_file:
-    - .env
     ports:
     - 8080:80


pengyin-shan:
> Well, in this scenario there is no more discovery app :) It is now only deployed at https://search.sdohplace.org (hence the new redirect from https://sdohplace.org/search), which is built from the new repository: https://github.com/healthyregions/sdohplace-data-discovery, not this one. So any issues with the discovery app should now be tracked in that repo.

Make sense! I'll transfer what I noticed in the new repo

bodom0015 (approved):
LGTM :+1: I have been clicking around a bit to test and have not noticed any majors errors

PR #709, opened 2025-12-01
fix: Research & Reports - parse Description from Body

### LINKED ISSUE #708: Research & Reports: Description content isn't displayed in CMS
Seems like for some reason the Description property isn't getting picked up as the body so it isn't parsed properly.

bodom0015:
## Problem
When using DecapCMS to edit an entry in Research & Reports, the Description field always shows as empty

If `body` is present in the MDX file, this body markdown text should be parsed as the description

Fixes #708 

Screenshot 2025-12-01 at 5 44 45 PM

## Approach
* fix: change `description` field to be named `body` instead, without changing the label in the UI

Screenshot 2025-12-01 at 5 55 16 PM

## How to Test
1. Navigate to /admin
2. On the left side, choose "Research Outputs"
    * You should see 2 entries listed here
3. Click on one of the entries
    * You should see the full editable metadata for this entry
4. Scroll all the way down
    * You should see the Description field is now populated by the MDX body

### DIFF
--- a/public/admin/config.yml
+++ b/public/admin/config.yml
@@ -479,7 +479,7 @@ collections:
         widget: string
         required: false
       - label: "Description"
-        name: "description"
+        name: "body"
         widget: "markdown"
 
   - name: "research-rabbit"


mradamcox (approved):
Perfect, good job figuring this out. Thanks!

PR #784, opened 2026-06-18
style: enforce char limit for CMS Post title/excerpt, display excerpt above post content

bodom0015:
## Problem
In DecapCMS, we've defined a collection of Posts (News, etc)

Posts have an `excerpt` field that provides a summary of the post content.
But this `excerpt` never appears on the full post itself.

The `excerpt` is displayed on the home page and has no char limit, and differences between length of the different excerpts can be visually distracting. 

## Approach
* style: enforce char limit for properties that are displayed on the homepage:
    * title = max limit of 60 chars
    * excerpt = max limit of 150 chars
* style: display `excerpt` above the Post content when showing the full Post
* feat: adjust `hint` for title + excerpt to prompt user about the max char limits

### CMS Admin
1. Navigate to https://deploy-preview-784--cheerful-treacle-913a24.netlify.app/admin/index.html and log in with your GitHub credentials
    * You should be logged into DecapCMS
2. On the left, choose Posts
3. At the top, click "+ Posts"
    * You should be brought to an empty editor that allows you to create a new post
    * You should see "Title" and "Excerpt" now have enhanced `hint` text displayed below each - these should mention the character limits.
4. Type a very long value into the `title` field (61+ characters)
    * You should see a red error message appear stating that title should be no longer than 60 characters
5. Type a very long value into the `excerpt` field (151+ characters)
    * You should see a red error message appear stating that excerpt should be no longer than 150 characters

### SDOH&Place Website
1. Navigate to https://deploy-preview-784--cheerful-treacle-913a24.netlify.app
2. Scroll down to the "Featured" section
    * You should see the font-weight of the title has been reduced from `1000` to `700`
3. Keep scrolling to the "News & Updates" section
4. On the right, click on "All News & Updates"
    * You should be brought to the News page, where all posts are listed
    * You should see that this page now includes the full text of the `excerpt` of each post
5. Choose a post from the list to see the full post
    * You should be brought to the full/rendered layout for this post
    * You should **not** see the excerpt in the header of the post


public/admin/config.yml
```diff
@@ -1,3 +1,4 @@
+local_backend: true
```
bodom0015: This line should not be included:
```suggestion
```

